PR #3, opened 2017-08-07
[#150015077] Fixing http auth and integrating it into ejabberd core

IRog:
In http_auth I had to modify the startup method. The supervision tree is slightly different but it shouldn't pose problems. I dunno why using the ejabberd_sup was causing issues, but I switched it to using direct start_links and there is still a supervisor in the chain, so it should be okay.

### DIFF
--- a/src/ejabberd_auth.erl
+++ b/src/ejabberd_auth.erl
@@ -524,61 +524,76 @@ db_set_password(User, Server, Password, Mod) ->
 
 db_get_password(User, Server, Mod) ->
     UseCache = use_cache(Mod, Server),
-    case erlang:function_exported(Mod, get_password, 2) of
-	false when UseCache ->
-	    ets_cache:lookup(?AUTH_CACHE, {User, Server});
-	false ->
-	    error;
-	true when UseCache ->
-	    ets_cache:lookup(
-	      ?AUTH_CACHE, {User, Server},
-	      fun() -> Mod:get_password(User, Server) end);
-	true ->
-	    Mod:get_password(User, Server)
+    case Mod == ejabberd_auth_http of
+      true ->
+        error;
+      false ->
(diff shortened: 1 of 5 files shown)


src/ejabberd_auth.erl
```diff
@@ -524,61 +524,76 @@ db_set_password(User, Server, Password, Mod) ->
...
-	    ets_cache:lookup(?AUTH_CACHE, {User, Server});
-	false ->
-	    error;
-	true when UseCache ->
-	    ets_cache:lookup(
-	      ?AUTH_CACHE, {User, Server},
-	      fun() -> Mod:get_password(User, Server) end);
-	true ->
-	    Mod:get_password(User, Server)
+    case Mod == ejabberd_auth_http of
```
myorg: Why was this needed?  Just wondering.

myorg: Looks like you're wrapping the preexisting logic in a case check to make sure that the mod is for the AUTH module.  Is that correct?

IRog: Yep. There was no other good way to fork the flow. Sometimes the check done in the parent case would explode if evaluated, so I needed to pre-empt that.

myorg: Got it.  Finished looking at the rest.  LGTM.

IRog: Thanks for looking. I'll wait for @thePhilGuy then merge.

src/ejabberd_auth_http.erl
```diff
@@ -0,0 +1,279 @@
+%%%----------------------------------------------------------------------
```
thePhilGuy: Did you change anything in the file below?

IRog: "In http_auth I had to modify the startup method."

thePhilGuy:
🎩

PR #7, opened 2017-08-29
updated README with instructions to copy ejabberd.yml

freester1:
@IRog


IRog:
Thanks. Was the ejabberctl stuff clear enough? If so 🌮

PR #8, opened 2017-09-25
[#151259942] this is a temp release with moar logs

IRog:


thePhilGuy:
For reference, we are trying to debug a P0 related to blocking users in which we're not sure which of these blocks is actually getting executed.

Talked about the log changes offline 🌮

IRog:
Thanks for additional context. I'm in head cut-off mode.

PR #10, opened 2017-09-26
still for the blocking issue. Trying to minimize the amount of changes

IRog:
This basically reverts 1 field change @thePhilGuy


thePhilGuy:
Best with a story ID, but 🌮

PR #12, opened 2017-10-20
[#151810061] Add tsung.xml 

freester1:
adding it here, I couldn't get great performance doing the custom xml requests 10% of the time, 
`error_send_no_ws_connection` being the main error. Also the custom requests aren't 100% since a lot of them require parsing the responses which I'm not sure if tsung can do out of the box

http://buildz.skillz.com/job/TS/job/Gatling/job/TsungStagingtests/114/HTML_Report/report.html

@IRog


IRog:
🎩

PR #13, opened 2017-10-27
[#152151840] Added open tcp sockets and erlang process count to metrics plus possible un-intended pushes from blocked users fix

IRog:
So this also has a merge from upstream so it's kinda huge.

Just check src/mod_metrics.erl and src/ejabberd_c2s.erl for my changes. I also should have made future merges easier. (https://github.com/skillz/ejabberd/pull/13/commits/b1a4a5c3e439f075caf768387520ad3c2aba046b)

Everything is still working fine locally and I tested the mod_metrics changes locally.

@freester1 
@thePhilGuy


freester1:
📖

PR #15, opened 2017-11-03
[#152552772] Updating code I put in before and think local test was working

IRog:
@freester1 
@thePhilGuy 

(for fixing notifications for friends from blocked users)


freester1:
🌮

PR #17, opened 2017-12-08
[#153525958] Updates to tsung for load testing

IRog:
@freester1 

I put it to 9. (I ran out of users in the csv, so when we get more we can go higher)


IRog:
I got to about 15% cpu usage on the 1 box it's hitting and doesn't seem anywhere close to capped.

IRog:
http://buildz.skillz.com/job/TS/job/Gatling/job/TsungStagingtests/156/HTML_Report/report.html

freester1:
🤖

PR #19, opened 2018-01-17
[#154293031] Our resources seem to be uuids so Im just kicking all a users resources so we dont need to specify it

IRog:
Also upped tsung from awhile ago
@mpope9


mpope9:
Looks good 🐘

IRog:
Will wait to confirm with TS for flow before merging

PR #21, opened 2018-04-02
[#155861133] Feature preserve messages

mpope9:
What this PR is accomplishing:

1)  *NOT* removing a room from the `muc_room` table the when it is destroyed (e.g. when last person leaves, destroyed through cmd line, etc.)
2)  Room, *is* removed from `muc_room` table and re-entered when a config change occurs (so we can change all of the existing rooms from persistent -> temporary).
3)  mod_mam's `archive` table's rows are never forgotten, in any case, ever.
4)  All rooms are stored in the DB, regardless of their state.
5)  Two scripts for starting/stopping the local server for ide development
6)  Steps in the README defining how to set up the basic ejabberd sql table for development.

@leejonaNEU 
@kkaminski


mix.lock
```diff
@@ -1,19 +0,0 @@
...
-  "jiffy": {:hex, :jiffy, "0.14.11", "919a87d491c5a6b5e3bbc27fafedc3a0761ca0b4c405394f121f582fd4e3f0e5", [:rebar3], [], "hexpm"},
-  "lager": {:hex, :lager, "3.4.2", "150b9a17b23ae6d3265cc10dc360747621cf217b7a22b8cddf03b2909dbf7aa5", [:rebar3], [{:goldrush, "0.1.9", [repo: "hexpm", hex: :goldrush, optional: false]}], "hexpm"},
-  "p1_oauth2": {:hex, :p1_oauth2, "0.6.1", "4e021250cc198c538b097393671a41e7cebf463c248980320e038fe0316eb56b", [:rebar3], [], "hexpm"},
-  "p1_utils": {:hex, :p1_utils, "1.0.9", "c33c230efbeb4dcc02911161e3cb1a93231a92df15e3fc97de655a9271a26d9f", [:rebar3], [], "hexpm"},
-  "stringprep": {:hex, :stringprep, "1.0.9", "9182ba39931cd1db528b8883cad0d63530abe2bf21835d26cec2f9af8bc00be0", [:rebar3], [{:p1_utils, "1.0.9", [repo: "hexpm", hex: :p1_utils, optional: false]}], "hexpm"},
-  "stun": {:hex, :stun, "1.0.14", "6dc2080c25a72f7087301dc7333c1ea7d27ea4d88efaa379fc2b5924f3b17006", [], [{:fast_tls, "1.0.15", [hex: :fast_tls, repo: "hexpm", optional: false]}, {:p1_utils, "1.0.9", [hex: :p1_utils, repo: "hexpm", optional: false]}], "hexpm"},
-  "xmpp": {:git, "https://github.com/processone/xmpp", "b8debe2790f836dcb6b2234a57565b2870b21677", []}}
```
kkaminski: what's going on here?

src/mod_muc.erl
```diff
@@ -607,7 +606,7 @@ get_rooms(ServerHost, Host) ->
     Mod = gen_mod:db_mod(LServer, ?MODULE),
     Mod:get_rooms(LServer, Host).
 
-load_permanent_rooms(Host, ServerHost, Access,
+load_all_rooms(Host, ServerHost, Access,
```
kkaminski: why rename this?

src/mod_muc.erl
```diff
@@ -168,7 +168,6 @@ restore_room(ServerHost, Host, Name) ->
 
 forget_room(ServerHost, Host, Name) ->
     LServer = jid:nameprep(ServerHost),
-    ejabberd_hooks:run(remove_room, LServer, [LServer, Name, Host]),
```
kkaminski: Maybe add a comment about how we removed this and why?

mix.lock
mpope9: Dependency resolution for elixir's `mix` utility.  I cleaned my directory.  Will add to the git ignore.

src/mod_muc.erl
mpope9: Because we are now loading temporary rooms too.

src/mod_muc_admin.erl
```diff
@@ -774,8 +774,7 @@ find_serverhost(Host, ServerHosts) ->
 act_on_room(destroy, {N, H, Pid}, SH) ->
     p1_fsm:send_all_state_event(
       Pid, {destroy, <<"Room destroyed by rooms_unused_destroy.">>}),
-    mod_muc:room_destroyed(H, N, Pid, SH),
-    mod_muc:forget_room(SH, H, N);
```
kkaminski: Isn't this actually the meat of the change?  not sure what's going on above, and again I'd put a comment in.

src/mod_muc.erl
mpope9: Its in the git history?  But can do

src/mod_muc_admin.erl
mpope9: Its all meat.  This is the part where we don't remove the room from mam, only from the muc_online_rooms in-memory table.

src/mod_muc_room.erl
```diff
@@ -3295,13 +3295,12 @@ change_config(Config, StateData) ->
...
-      {true, false} ->
-	  mod_muc:forget_room(NSD#state.server_host,
-			      NSD#state.host, NSD#state.room);
-      {false, false} -> ok
+      {_, _} ->
+	      mod_muc:forget_room(NSD#state.server_host,
+		      NSD#state.host, NSD#state.room),
+	      mod_muc:store_room(NSD#state.server_host,
+		      NSD#state.host, NSD#state.room, make_opts(NSD));
+	    {false, false} -> ok
```
kkaminski: alignment seems off here.... and I'm not sure if this fallback is ever going to happen?

src/mod_muc_room.erl
```diff
@@ -4027,12 +4020,11 @@ element_size(El) ->
...
-    if (StateData#state.config)#config.persistent ->
-	    mod_muc:store_room(StateData#state.server_host,
+    case mod_muc:store_room(StateData#state.server_host,
 			       StateData#state.host, StateData#state.room,
-			       make_opts(StateData));
-      true ->
-        ok
+			       make_opts(StateData)) of
+       {true} ->
+          ok
```
kkaminski: what's going on here?

src/mod_muc.erl
kkaminski: Yeah I'm just thinking about whenever we go to upgrade ejabberd, it would be nice to have some context for some of this stuff in the code.

kkaminski:
Feels like the config changes should be in a separate PR in case there's problems - that one is making me more nervous then the archive history preservation.

Also where did you land as far as unit tests & this stuff?

kkaminski:
Were you able to test locally at all?

src/mod_muc_room.erl
mpope9: Switched this from the if statement that only allows the storage of the room when persistent to allowing storage regardless of state and then pattern matching to a case statement to return what it was expecting before.  Defaults to `error`, I believe

src/mod_muc.erl
mpope9: Will add.

mpope9:
I couldn't test things locally without these changes.  But since this looks like it took a short amount of time, I can spend more effort on setting up sql.

mpope9:
@kkaminski No unit tests.  I was going to land this on QA, as I can't seem to be able to get the SQL working.  Once I see that everything looks good on QA I was going to go back and add to the unit tests.

src/mod_muc_room.erl
mpope9: Fixed the indentation.  And its possible, if it was never a persistent room in the past or if we're not setting it then it won't affect the temporary room.

src/mod_muc.erl
kkaminski: Maybe this is a dumb suggestion since we'd probably be doing it a lot - feel free to ignore.

src/mod_muc.erl
kkaminski: Oh because we're not clearing from the archive table, it loads all the room history automatically?

src/mod_muc_room.erl
```diff
@@ -3295,13 +3295,12 @@ change_config(Config, StateData) ->
...
 	  Config#config.persistent}
 	of
-      {_, true} ->
-	  mod_muc:store_room(NSD#state.server_host,
-			     NSD#state.host, NSD#state.room, make_opts(NSD));
-      {true, false} ->
-	  mod_muc:forget_room(NSD#state.server_host,
-			      NSD#state.host, NSD#state.room);
-      {false, false} -> ok
+      {_, _} ->
```
leejonaNEU: is it really any matcher? If thats the case feels like a case statement is redundant

src/mod_muc_room.erl
```diff
@@ -3295,13 +3295,12 @@ change_config(Config, StateData) ->
...
+        mod_muc:forget_room(NSD#state.server_host,
+          NSD#state.host, NSD#state.room),
+	      mod_muc:store_room(NSD#state.server_host,
+          NSD#state.host, NSD#state.room, make_opts(NSD));
+        {false, false} -> ok
```
leejonaNEU: wouldn't `{_, _}` catch everything before you're able to reach `{false, false}`? Or am I reading this wrong

src/mod_muc_room.erl
leejonaNEU: nit but seems like you're just using the case as an if statement? Seems unnecessarily clunky, why not just change the condition of the if statement?

src/mod_muc.erl
mpope9: Oh hmmm I think this call should be removed, actually

src/mod_muc_room.erl
mpope9: Whoops, wrong order, you're right

src/mod_muc_room.erl
mpope9: Different paradigms for erlang.  http://erlang.org/pipermail/erlang-questions/2008-March/033673.html

src/mod_muc_room.erl
mpope9: I have 2 specific cases where I want things to happen.  This makes it explicit.

.gitignore
```diff
@@ -47,3 +47,8 @@ XmppAddr.hrl
 /_build
 /mnesiadb
 /.rebar
+
+.idea/
+ejabberd.iml
+*log/
+mix.lock
```
kkaminski: Hm I don't see us doing this elsewhere - you may have a version mismatch, but I guess we can leave this.

README
```diff
@@ -29,6 +29,24 @@ make && make install
...
+And make an ejabberd user and database in our local mysql docker.
+0) 
+```
+docker ps | grep my
+docker exec -it <hash_from_above> bash
+mysql
+```
+1) Create the ejabberd user and use '%' for the host.
+2) Create the ejabberd table
+2.5) Exit the container
```
kkaminski: nit, renumber and start at 1 instead of 0

mix.lock
```diff
@@ -2,18 +2,19 @@
...
   "p1_utils": {:hex, :p1_utils, "1.0.9", "c33c230efbeb4dcc02911161e3cb1a93231a92df15e3fc97de655a9271a26d9f", [:rebar3], [], "hexpm"},
   "stringprep": {:hex, :stringprep, "1.0.9", "9182ba39931cd1db528b8883cad0d63530abe2bf21835d26cec2f9af8bc00be0", [:rebar3], [{:p1_utils, "1.0.9", [repo: "hexpm", hex: :p1_utils, optional: false]}], "hexpm"},
-  "stun": {:hex, :stun, "1.0.14", "6dc2080c25a72f7087301dc7333c1ea7d27ea4d88efaa379fc2b5924f3b17006", [], [{:fast_tls, "1.0.15", [hex: :fast_tls, repo: "hexpm", optional: false]}, {:p1_utils, "1.0.9", [hex: :p1_utils, repo: "hexpm", optional: false]}], "hexpm"},
-  "xmpp": {:git, "https://github.com/processone/xmpp", "b8debe2790f836dcb6b2234a57565b2870b21677", []}}
+  "stun": {:hex, :stun, "1.0.14", "6dc2080c25a72f7087301dc7333c1ea7d27ea4d88efaa379fc2b5924f3b17006", [:rebar3], [{:fast_tls, "1.0.15", [hex: :fast_tls, repo: "hexpm", optional: false]}, {:p1_utils, "1.0.9", [hex: :p1_utils, repo: "hexpm", optional: false]}], "hexpm"},
+  "xmpp": {:git, "https://github.com/processone/xmpp", "fab6269a36cbd103f203ad75b7097650987b7a85", [tag: "1.1.14"]}}
```
kkaminski: Can you revert changes to this?

src/mod_muc.erl
mpope9: Removed.  Now they're not loaded into memory upon initialization.

kkaminski: why?  Also isn't this part of #4 (in #feature-chat)?  I thought we were doing a separate PR for that.

README
mpope9: Okay

mix.lock
mpope9: reverted

kkaminski:
I don't have anything major here, just trying to understand the changes moreso then anything. It sounds like we want to rush this to QA w/o local test or testing so 🤞

leejonaNEU:
👍

PR #22, opened 2018-04-04
[#155861133] Feature room deletion script

mpope9:
Python script to create 5k rooms.  Python script to delete group messages and DMs

@aghchan 
@RuoanJi

### DIFF
--- a/delete_rooms.py
+++ b/delete_rooms.py
@@ -3,6 +3,7 @@
 import subprocess
 import os
 
+# JID is userId1-userId2@conference.chat.skillz.com  Global rooms have no '-', so they're unaffected.
 # Load all rooms with a - (dms and group messages) 
 from_ctl = subprocess.Popen(['./sbin/ejabberdctl', 'muc_online_rooms', 'global'], stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


delete_rooms.py
```diff
@@ -0,0 +1,19 @@
...
+from_ctl = subprocess.Popen(['./sbin/ejabberdctl', 'muc_online_rooms', 'global'], stdout=subprocess.PIPE)
+all_rooms = subprocess.Popen(['grep', '-'], stdin = from_ctl.stdout, stdout = subprocess.PIPE)
+from_ctl.stdout.close()
+
+muc_rooms, stderr = all_rooms.communicate()
+
+# Loop through all muc rooms and delete them.
+for muc in muc_rooms.split(os.linesep):
+    name_service = muc.replace('@', ' ')
+    name_service_list = name_service.split(' ')
```
aghchan: can you comment the format of the mucJid and does the name contain more than one `@`? if not you can just split on the `@`

mpope9: Oh good call, will change

aghchan:
👌

RuoanJi:
the script looks good to me 👍 

not sure if `ejabberdctl` has some kind of limitations if you run commands too many too fast

mpope9:
You were right @RuoanJi, had to add sleep.

PR #23, opened 2018-04-05
[#155861133] Fixing the ownership issue

mpope9:
Now we have members and owners.  NO ADMINS.  

@aghchan 
@RuoanJi


aghchan:
So in the event that the member and owner leaves and the room is deleted, this allows the member to recreate the room right?

mpope9:
Yeah, but the room getting deleted part isn't complimented yet.  this is for when I manually delete a room both users can rejoin

aghchan:
💬

PR #28, opened 2018-04-06
[#156577214] Assigning the state data that was changed

mpope9:
Didn't assign the variable.

@aghchan 
@RuoanJi


aghchan:
🆗 we goofed

PR #29, opened 2018-04-06
[#156577214] Adding ownership upon mucsub request.

mpope9:
Didn't remember we used muc sub.  So setting ownership upon receiving a MUC SUB

@aghchan 
@RuoanJi 
@kkaminski


aghchan:
⭕️

PR #30, opened 2018-04-10
[#155861133] Ownership binding

mpope9:
Attempting to bind to the StateData variables to see if its saved in the DB after this.

@aghchan 
@kkaminski


aghchan:
🆗

PR #31, opened 2018-04-10
[#156664850] Save Muc Room in DB When Running destroy_room

aghchan:
Removing this so we do not run the forget_room which deletes the room from the DB

@mpope9 
@mattfoley


mpope9:
LGTM 🐘

MattFoley:
👍

PR #32, opened 2018-04-10
[#156665132] Admin can create rooms

mpope9:
Removing all instances of changing affiliation.  Now just allowing admins to do owner things where the error message appears.

@aghchan


aghchan:
👌

PR #34, opened 2018-04-10
[#156666525] Removing the remove_subscription call

mpope9:
Removing the call to the `remove_subscriptions` from `change_config`, which is called in `set_config` which is called in several relevant places.

@aghchan


aghchan:
👌

PR #38, opened 2018-04-12
[#156666629] Create Room When Joining if it Does Not Exist

aghchan:
Reverted changes yesterday as it caused chat to be broken and found in the logs that the methods were undefined, causing the server to crash. Added those methods to the export so mod_muc_room can use these methods from mod_muc

@mpope9 
@MattFoley


mpope9:
🐘 Looks better!

PR #39, opened 2018-04-12
[#156666629] Register Online Room After Creation

aghchan:
After the last PR, saw that the `RMod:register_online_room` crashed the server - going to use the exposed one in `mod_muc` instead

@mpope9


mpope9:
🐘

PR #41, opened 2018-04-13
[#156666629] Added Default Case For Not Creating Room When Joining

aghchan:
Handling case when not creating a room when joining

@mpope9 
@MattFoley


mpope9:
🐘

PR #42, opened 2018-04-13
[#156666629] Send Signal that Room was Created

aghchan:
Looking at `mod_muc`, after it registers the room it routes the packet to signal that the room was created, so will do the same here.

@mpope9 
@MattFoley


mpope9:
🔌 One line at a time.

PR #51, opened 2018-04-16
[#155907078] Reload Messages from MAM on Recreated Rooms

aghchan:
Querying DB when history is empty, and sending event that history has been retrieved. Added logs while I'm at it.

@mpope9 
@MattFoley


mpope9:
🐘 If you remove the limit max

PR #54, opened 2018-04-17
[#156709543] Feature sql subscription list

mpope9:
This creates a table for subscriptions, instead of pulling them from memory.

`name` is the JID of the client.  Index balanced by a BTREE, following the conventions found in the other tables in the Ejabberd db.

`subscription` is a single room that the user is subscribed to.

@aghchan 
@kkaminski

### DIFF
--- a/migrations/2018-04-17-subscribers-table.sql
+++ b/migrations/2018-04-17-subscribers-table.sql
@@ -9,4 +9,4 @@ CREATE TABLE subscriptions (
 ) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 CREATE INDEX i_subscriptions_name_index USING BTREE on subscriptions(name(25));
 CREATE INDEX i_subscriptions_subscription_index USING BTREE on subscriptions(subscription(25));
-CREATE UNIQUE INDEX i_subscriptions_created_at_index on subscriptions(created_at);
+CREATE INDEX i_subscriptions_created_at_index on subscriptions(created_at);
--- a/migrations/2018-04-17-subscribers-table.sql
+++ b/migrations/2018-04-17-subscribers-table.sql
@@ -1,5 +1,7 @@
 -- This file creates a new table, subscriptions, in the ejabberd DB.
 
+USE ejabberd;
+
 CREATE TABLE subscriptions (
     name text NOT NULL,
     subscriber_list mediumtext NOT NULL,
--- a/migrations/2018-04-17-subscribers-table.sql
+++ b/migrations/2018-04-17-subscribers-table.sql
@@ -7,5 +7,6 @@ CREATE TABLE subscriptions (
     subscription text NOT NULL,
     created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
 ) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
-CREATE UNIQUE INDEX i_subscriptions_subscription_index USING BTREE on subscriptions(subscription(25));
+CREATE INDEX i_subscriptions_name_index USING BTREE on subscriptions(name(25));
+CREATE INDEX i_subscriptions_subscription_index USING BTREE on subscriptions(subscription(25));
 CREATE UNIQUE INDEX i_subscriptions_created_at_index on subscriptions(created_at);
--- a/migrations/2018-04-17-subscribers-table.sql
+++ b/migrations/2018-04-17-subscribers-table.sql
@@ -4,8 +4,8 @@ USE ejabberd;
 
 CREATE TABLE subscriptions (
     name text NOT NULL,
-    subscriber_list mediumtext NOT NULL,
+    subscription text NOT NULL,
     created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
 ) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
-CREATE UNIQUE INDEX i_subscriber_name_index USING BTREE on subscriptions(name(25));
-CREATE UNIQUE INDEX i_subscriber_created_at_index on subscriptions(created_at);
+CREATE UNIQUE INDEX i_subscriptions_subscription_index USING BTREE on subscriptions(subscription(25));
+CREATE UNIQUE INDEX i_subscriptions_created_at_index on subscriptions(created_at);
--- a/sql/2018-04-17-subscribers-table.sql
+++ b/sql/2018-04-17-subscribers-table.sql
@@ -1,14 +1,12 @@
--- This file creates a new table, subscriptions, in the ejabberd DB.
+-- This file creates a new table, subscription, in the ejabberd DB.
 
 USE ejabberd;
 
 CREATE TABLE subscription (
-    name varchar(50) NOT NULL,
-    subscription varchar(255) NOT NULL,
+    jid varchar(50) NOT NULL,
+    room varchar(255) NOT NULL,
     last_updated timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
 ) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 CREATE INDEX i_subscription_name_index USING BTREE on subscription(name(25));
-CREATE INDEX i_subscription_subscription_index USING BTREE on subscription(subscription(25));
-CREATE INDEX i_subscription_last_updated_index on subscription(last_updated);
 CREATE INDEX i_subscription_created_at_index on subscription(created_at);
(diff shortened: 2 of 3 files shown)


aghchan:
👌

migrations/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
+-- This file creates a new table, subscriptions, in the ejabberd DB.
+
+USE ejabberd;
+
+CREATE TABLE subscriptions (
```
kkaminski: Nit, but our standard is not to pluralize table names in TS and I see table names like `archive` here, so I'd drop the `s`

migrations/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
...
+    name text NOT NULL,
+    subscription text NOT NULL,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
```
kkaminski: how about `last_updated`?

migrations/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
...
+) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+CREATE INDEX i_subscriptions_name_index USING BTREE on subscriptions(name(25));
+CREATE INDEX i_subscriptions_subscription_index USING BTREE on subscriptions(subscription(25));
+CREATE UNIQUE INDEX i_subscriptions_created_at_index on subscriptions(created_at);
```
kkaminski: why is this unique?

kkaminski:
why isn't this in the `sql` directory with the other schema definitions?

migrations/2018-04-17-subscribers-table.sql
kkaminski: But if it's easier to leave this this way for existing code, then just ignore me

migrations/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
+-- This file creates a new table, subscriptions, in the ejabberd DB.
+
+USE ejabberd;
+
+CREATE TABLE subscriptions (
+    name text NOT NULL,
```
kkaminski: generally best practice to define varchars with fixed lengths (if you know what the max size could be).

mpope9: Will change to varchar(50) since its the concatenation of 2 ids.

mpope9: Should I change the other text as well?

mpope9:
@kkaminski Moved to the sql directory, didn't know it existed!

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
...
+CREATE TABLE subscription (
+    name varchar(50) NOT NULL,
+    subscription varchar(255) NOT NULL,
+    last_updated timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
+) ENGINE=InnoDB CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+CREATE INDEX i_subscription_name_index USING BTREE on subscription(name(25));
+CREATE INDEX i_subscription_subscription_index USING BTREE on subscription(subscription(25));
+CREATE INDEX i_subscription_last_updated_index on subscription(last_updated);
+CREATE INDEX i_subscription_created_at_index on subscription(created_at);
```
kkaminski: @mpope9 indexes can be expensive if they're not used.  You mentioned @zgarbowitz suggested these, I think he should take a quick look so I'll DM him.

zgarbowitz: Oh this must've been a misunderstanding- I think we should have indexes on name and created_at here; I don't think we need them on subscription or last_updated

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
+-- This file creates a new table, subscriptions, in the ejabberd DB.
```
kkaminski: * `subscription`

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
...
+
+USE ejabberd;
+
+CREATE TABLE subscription (
```
kkaminski: No id?

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
...
+    name varchar(50) NOT NULL,
+    subscription varchar(255) NOT NULL,
+    last_updated timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
```
kkaminski: Is there a primary key in this table?

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
+-- This file creates a new table, subscriptions, in the ejabberd DB.
+
+USE ejabberd;
+
+CREATE TABLE subscription (
+    name varchar(50) NOT NULL,
+    subscription varchar(255) NOT NULL,
```
kkaminski: Sooo I'm actually not sure what a `subscription` is or how this table is going to be used, but `subscription.subscription` is kind of bad form - is there any way to make that more identifyable

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,14 @@
+-- This file creates a new table, subscriptions, in the ejabberd DB.
+
+USE ejabberd;
+
+CREATE TABLE subscription (
+    name varchar(50) NOT NULL,
```
kkaminski: is this a room name?

sql/2018-04-17-subscribers-table.sql
mpope9: Will change.

sql/2018-04-17-subscribers-table.sql
mpope9: None of the other ejabberd tables have id, and I was following that convention.  Is that okay?

sql/2018-04-17-subscribers-table.sql
mpope9: No, we want repeat for each column

sql/2018-04-17-subscribers-table.sql
mpope9: I was going to have `subscriptions.subscription`.  I could just change it to `room`

sql/2018-04-17-subscribers-table.sql
mpope9: Its the jid of the client

mpope9: changed to `jid

sql/2018-04-17-subscribers-table.sql
mpope9: changed to `room`

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
+-- This file creates a new table, subscription, in the ejabberd DB.
+
+USE ejabberd;
+
+CREATE TABLE subscription (
+    jid varchar(50) NOT NULL,
```
kkaminski: Should we have a unique constraint on jid/room or do you expect there to be dups intentionally?

mpope9: Dups are intentional.  A user can have more than one subscription

sql/2018-04-17-subscribers-table.sql
```diff
@@ -0,0 +1,12 @@
...
+    room varchar(255) NOT NULL,
+    last_updated timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP
```
kkaminski: ts standard is `date_created` unless there's a reason for this to be different

mpope9: Ejabberd standard has `created_at` and I was following that for most of these

sql/2018-04-17-subscribers-table.sql
kkaminski: To the same room?

kkaminski: (I'm suggesting a unique constraint on `jid & room` although now that I'm thinking it, I wonder if that's a performance concern given that these are strings)

sql/2018-04-17-subscribers-table.sql
kkaminski: kk

kkaminski:
🎊 as long as we handle the unique constraint programmatically

sql/2018-04-17-subscribers-table.sql
mpope9: Here is the query that I'm using

```        ejabberd_sql:sql_query_t(
                  ?SQL("insert into subscription (jid, room) "
                       "select %(SJID)s, %(SRoomJID)s from DUAL "
                       "where not exists ("
                       "   select 1 from subscription "
                       "   where jid=%(SJID)s AND room=%(SRoomJID)s);"))```

Example:
user 9 and 10 make room 9-10
user 9 and 11 make room 9-11

The db will have
jid      | room
9        | 9-10
10      | 9-10
9        | 9-11
11       | 9-11

PR #55, opened 2018-04-19
[#156709610] Feature add subscribers

mpope9:
This PR is the start of persistent subscription lists.

Originally the plan was to save these in a blob, but then the issue arose that each time a user logged in we'd have to loop through the blob to see if appending was needed.  I have chosen to go with the multi row approach, because of this.

The SQL query ensures that only one entry per subscription.

Some background: each muc room is a process.  The 'muc manager' is its own process.  When a subscription request comes it, I create a hook for the muc manager.  The muc room calls the hook and the manager stores the result in the database.

Only implemented the sql, added error returns if other databases are attempted to be used.

### Testing

Create a room with a friend.
Saw two rows were created.
Stopped the server.
Rows still exist.
Started the server.
Reopened the muc room.
No extra entry.

@aghchan 
@kkaminski

### DIFF
--- a/src/mod_muc_sql.erl
+++ b/src/mod_muc_sql.erl
@@ -65,8 +65,6 @@ db_subscribe(LServer, LBareJID, LBareRoomJID) ->
     SRoomJID = jid:encode(LBareRoomJID),
     F = fun () ->
         ejabberd_sql:sql_query_t(
-                  %%?SQL("insert into subscriptions (name, subscription) values( "
-                  %%     "%(SJID)s, %(Room)s);"))
                   ?SQL("insert into subscriptions (name, subscription) "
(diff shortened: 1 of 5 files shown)


src/mod_muc_sql.erl
```diff
@@ -56,6 +57,22 @@ init(Host, Opts) ->
 	    ok
     end.
 
+%% Select the JID from the DB.  If it does not exist, create the row.  Otherwise,
+%% concat the current list of subscriptions with the new one.  Lets hope this doesn't
+%% get too long, it is a medium text after all.
+db_subscribe(LServer, LBareJID, LBareRoomJID) ->
```
aghchan: should update the comment to the multi-row approach

PR #56, opened 2018-04-20
[#156709589] Subscriptions retrieved and displayed

mpope9:
Subscriptions are now persistent and no longer reliant on the muc room being in memory.  A bug was created as a side effect where all of the subscribed rooms that a user is subscribed to are brought back into memory, possibly due to a different SDK call.  This exacerbates the issue of too many muc rooms residing in memory.  Will file a separate bug to fix this.

Much like the other PR the mneisa and riak implementations are not going to be completed, as they will not be used.

Tested using multiple registered muc rooms, including 1x1 and 1x1x1.  

@RuoanJi 
@MattFoley


RuoanJi:
LGTM 👍

PR #57, opened 2018-04-23
[#156968485] Closes the room if no occupants exist

mpope9:
Now when we retrieve muc subscribers the room will not persist.  However, it is loaded into memory for a short amount of time.

This potentially fixes the other bug where muc rooms are persisted after the last user leaves.

@RuoanJi 
@MattFoley


RuoanJi:
👍 

Talked offline, I am not fully understanding the state machine behind. The state transition and the new state look good.

PR #58, opened 2018-04-27
[#156864336] Feature reload mam

mpope9:
archive table is queried upon muc_room_initialization and the archived messages are loaded into the muc room history.  Right now querying the last 50 messages as a hard coded value.

@aghchan 
@MattFoley

### DIFF
--- a/src/mod_mam.erl
+++ b/src/mod_mam.erl
@@ -39,6 +39,8 @@
 	 get_commands_spec/0, msg_to_el/4, get_room_config/4, set_room_option/3,
 	 offline_message/1, export/1]).
 
+-export([get_room_history/3]).
+
 -include("xmpp.hrl").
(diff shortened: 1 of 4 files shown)


src/mod_mam.erl
```diff
@@ -37,7 +37,9 @@
 	 remove_user/2, remove_room/3, mod_opt_type/1, muc_process_iq/2,
 	 muc_filter_message/3, message_is_archived/3, delete_old_messages/2,
 	 get_commands_spec/0, msg_to_el/4, get_room_config/4, set_room_option/3,
-	 offline_message/1, export/1]).
+	 offline_message/1, send/4, export/1]).
```
aghchan: extra export for send here unused

mpope9: Fixed

PR #60, opened 2018-05-02
[#157237109] Fixing the messages that are retrieved

mpope9:
Tested with a room with more than 50 messages.  The 50 MOST RECENT messages are retrieved in current time sorted order.

@aghchan 
@MattFoley


aghchan:
🆗

PR #64, opened 2018-05-03
[#157257017] Fixing timestamp format

mpope9:
I misunderstood the timestamp format.  This should fix it.

@aghchan 
@MattFoley


aghchan:
👌

PR #67, opened 2018-09-07
[#159511907] Sending ejabberd messages through the offline flow

mpope9:
Instead of directly sending the message, now we shall push it through the offline message flow.

Can't test this locally, obviously.

@kkaminski 
@RuoanJi


RuoanJi:
Let's give this a try 👍

david-thornton717:
Ok 🌜

PR #69, opened 2018-10-05
[#160483755] bug message type offline message flow check

mpope9:
If the message type is included in the list of message types then only send to the user(s) who didn't send the message.

@aghchan 
@kkaminski

### DIFF
--- a/include/message_type.hrl
+++ b/include/message_type.hrl
@@ -15,3 +15,5 @@
          ?VsFriendsMessageType, 
          ?NudgeMessageType
         ]).
+
+-define(SdkElementsPosition, 3).
--- a/include/message_type.hrl
+++ b/include/message_type.hrl
@@ -1,6 +1,6 @@
 %%------------------------------------------------------------------------
 %%
-%% This file contains skillz message types that we shouldn't include in the 
+%% This file contains message types that we shouldn't include in the 
 %% offline message flow.
 %%
 %% Also contains various constant that we need to access these items.
--- a/include/message_type.hrl
+++ b/include/message_type.hrl
@@ -3,7 +3,7 @@
 %% This file contains message types that we shouldn't include in the 
 %% offline message flow.
 %%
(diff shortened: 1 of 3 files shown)


kkaminski:
`Feature`?

include/message_type.hrl
```diff
@@ -0,0 +1,17 @@
+%%------------------------------------------------------------------------
+%%
+%% This file contains message types that we shouldn't include in the 
+%% offline message flow.
+%%
+%% Also contains various constant that we need to access these items.
```
kkaminski: not sure what this sentence means?

kkaminski:
no tests?

include/message_type.hrl
mpope9: Whoops, meant to put that hard-coded `3` value below in here.

mpope9:
No tests because the ejabberd tests fail and its quite arduous to set it up, much like before.  We are going to bring this up to processone when we have the call, and we have a story to go back and add tests for all of the code that we have changed.

aghchan:
👌

src/mod_muc_room.erl
```diff
@@ -4071,6 +4073,35 @@ is_privacy_allow(Packet) ->
     LServer = To#jid.server,
     allow == ejabberd_hooks:run_fold(privacy_check_packet, LServer, allow, [To, Packet, in]).
 
+-spec inspect_sdk_xmlels(binary(), map(), list()) -> list().
```
satishv: Can we add more doco about specs?

Can you add a pull request to community edition too? so that you can get feedback from the community? especially we do not have in house experts on this, yet...

https://github.com/processone/ejabberd/blob/master/src/mod_muc_room.erl

mpope9: Talked offline, will make a forum post

kkaminski:
👍

PR #70, opened 2018-10-09
[#161064486] Update docker build

balusarakesh:
overview:
changes to Dockerfile to make it work


Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM debian:jessie-slim
-MAINTAINER Rafael Römhild <rafael@roemhild.de>
+FROM phusion/baseimage:0.9.22
 
-ENV EJABBERD_BRANCH=17.08 \
+ENV EJABBERD_BRANCH=master \
```
thePhilGuy: @balusarakesh @mpope9 I think we want the development branch here, since this docker container will be used in the context of testing locally.

balusarakesh: This is the third-party ejabberd which we download and compile for our usecase. It's not ideal to use development branch from other repos.

balusarakesh: please refer to https://github.com/skillz/ejabberd/blob/b9110370c789981272cb18ca3123dc44a1518d43/Dockerfile#L64

thePhilGuy: Oh then good thing I asked this question, we do not want to compile the 3rd party ejabberd.
Instead, we want to compile THIS version, on the development branch.

So instead of cloning here: https://github.com/skillz/ejabberd/blob/b9110370c789981272cb18ca3123dc44a1518d43/Dockerfile#L64
This dockerfile should have access to this repository ( github.com/skillz/ejabberd ) and `make` that

Dockerfile
```diff
@@ -1,113 +1,44 @@
...
+    zlib1g-dev \
+    libpng-dev \
+    libexpat-dev \
+    git \
+    automake \
+    libssl-dev \
+    g++
+
+# install ejabberd
+ARG EJABBERD_HOME=/root/ejabberd
```
thePhilGuy: Can we change this to `/opt/ejabberd` to replicate something closer to the actual deployments.

Dockerfile
```diff
@@ -1,113 +1,44 @@
...
+ARG MY_EJABBERD_HOME=/root/my-ejabberd
+
+RUN mkdir $EJABBERD_HOME
+RUN git clone https://github.com/skillz/ejabberd.git --branch development $EJABBERD_HOME
+COPY ejabberd.yml $EJABBERD_HOME/config/ejabberd.yml
+RUN cd $EJABBERD_HOME && ./autogen.sh
+RUN cd $EJABBERD_HOME && ./configure --prefix=$MY_EJABBERD_HOME --enable-elixir
+RUN cd $EJABBERD_HOME && make && make install
+COPY ejabberd.yml $MY_EJABBERD_HOME/etc/ejabberd/ejabberd.yml
```
thePhilGuy: Why is there also a $MY_EJABBERD_HOME ?

Is this to separate the compiled location from the source directory?
If that's the case then the compiled location should be `/opt/ejabberd`

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+## You can define one or several, for example:
+## hosts:
+##   - "example.net"
+##   - "example.com"
+##   - "example.org"
+##
+hosts:
+  - "dev-ejabberd"
+  - "localhost"
+  - "192.168.22.37"
```
thePhilGuy: What is this third IP ? 192.168.22... reminds me of our in-office addresses.
Is this necessary when running in docker? If so this will only work for whoever has 192.168.22.37

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+#
+# auth_method: Method used to authenticate the users.
+# The default method is the internal.
+# If you want to use a different method,
+# comment this line and enable the correct ones.
+#
+
+auth_method: http
+auth_opts:
+  host: "http://localhost:9100"
```
thePhilGuy: Does this work with `dev-ts` ? it'd be good to be explicit

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+
+auth_password_format: plain
+disable_sasl_mechanisms:
+  - "digest-md5"
+  - "scram-sha-1"
+
+default_db: sql
+
+sql_type: mysql
+sql_server: "127.0.0.1"
```
thePhilGuy: dev-mysql ?

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+  ## You can put here as many accounts as you want.
+  ##
+  admin:
+    user:
+      - "admin": "chat.dev.skillz.com"
+      - "admin": "localhost"
+      - "admin": "192.168.22.37"
+      - "skillz-cas": "chat.dev.skillz.com"
+      - "skillz-cas": "localhost"
+      - "skillz-cas": "192.168.22.37"
```
thePhilGuy: ditto on specific IP, should be taken out or dynamic

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+      #         - ip: "127.0.0.1/8"
+      #         - acl: admin
+    what:
+      - "send_stanza"
+      - "kick_session"
+  "public commands":
+    who:
+      - ip: "127.0.0.1/8"
+    what:
+      - "status"
```
thePhilGuy: Looks like this is allowed to anyone so we can use it as a healthcheck

Dockerfile
```diff
@@ -1,113 +1,44 @@
...
+
```
thePhilGuy: So I just realized it doesn't look like this installs modules at all. Wouldn't that be a problem for running/testing ejabberd?
This is the relevant section in the apt build script: https://github.com/skillz/deploy/blob/fbc455636f021da44fd827439f45d04a5038d8c6/conf/apt-repositories/build/ejabberd-dev.sh#L58

If ejabberd can start without those modules and you've used up all your points we can create another story. If not this is a blocker

mpope9: Yes, adding and installing the modules is required.  Completely glossed this over, can be added later though

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+##
+## hosts: Domains served by ejabberd.
+## You can define one or several, for example:
+## hosts:
+##   - "example.net"
+##   - "example.com"
+##   - "example.org"
+##
+hosts:
+  - "dev-ejabberd"
```
mpope9: can you also add `chat.dev.skillz.com`

ejabberd.yml
```diff
@@ -0,0 +1,510 @@
...
+###.   ====================
+###'   ACCESS CONTROL LISTS
+acl:
+  ##
+  ## The 'admin' ACL grants administrative privileges to XMPP accounts.
+  ## You can put here as many accounts as you want.
+  ##
+  admin:
+    user:
+      - "admin": "chat.dev.skillz.com"
```
mpope9: also add `admin` and `skillz-cas` for `dev-ejabberd`

balusarakesh: done

ejabberd.yml
balusarakesh: done

mpope9:
🐘 Once Philippe's comments are implemented.

Dockerfile
mpope9: Actually pushes do not work locally.  So this isn't really needed after all

thePhilGuy: Ok then 👍

mpope9:
🐘

PR #71, opened 2018-10-17
[#160483755] Fixing bad call to offline message

mpope9:
I did not include the third argument to this call.  See the messages going to the spool table now locally.

@aghchan 
@kkaminski


aghchan:
looks like it was already there before the previous change so 👍

bhickey-skillz:
👍

PR #77, opened 2019-02-27
[ES-950] Offline messgaes sent to the correct user for send_stanza

mpope9:
OK, I forgot under the hood we actually don't send to the user we include, but to the opposite user.  I guess I could make a PR in the server-api repo instead of this one, but I don't know if that will break more stuff.  It looks like the correct user is inserted in the offline message table.

@Tdavis22 
@zgarbowitz


zgarbowitz:
👍

PR #78, opened 2019-02-27
[ES-1437] Moving killing the muc from subscriptions to presence

mpope9:
This should fix the broader memory leak.  Basically reusing what I wrote here, https://github.com/skillz/ejabberd/pull/57/files.  Also removing the code that isn't getting hit by the sdk.

I can't get the SOCKS proxy working anymore...so I'm not able to test locally

@Tdavis22 
@zgarbowitz


Tdavis22:
👍 But I would be weary of Changing any of the Ejabberd code base especially when we plan to upgrade it soon with Process 1 patch

zgarbowitz:
👍

PR #79, opened 2019-02-28
[ES-1437] Kill mucs higher up in the call chain

mpope9:
Made a wrapper function that kills mucs higher up in the call chain.

Please note that I can't test this due to lack of local proxying, and there are 3 other functions where this issue can be happening.  If this issue persists, I can make another PR for the different functions.

This is because force closing the sdk re-opens rooms.

Left a note to remove this and add a module that cleans mucs periodically. 

@zgarbowitz 
@matlegit

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -156,8 +156,12 @@ init([Host, ServerHost, Access, Room, HistorySize, RoomShaper, Opts, QueueType])
     add_to_log(room_existence, started, NewState),
     {ok, normal_state, NewState}.
 
+%% This is a top level call to normal_state to kill the muc room during any event.
+normal_state(Event, StateData) ->
+	{_, _, NewStateData} = normal_state(Event, StateData),
+	close_room_without_occupants(NewStateData);
 normal_state({route, <<"">>,
-	      #message{from = From, type = Type, lang = Lang} = Packet},
+              #message{from = From, type = Type, lang = Lang} = Packet},
 	     StateData) ->
     case is_user_online(From, StateData) orelse
 	is_subscriber(From, StateData) orelse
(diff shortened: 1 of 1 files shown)


src/mod_muc_room.erl
```diff
@@ -156,9 +156,14 @@ init([Host, ServerHost, Access, Room, HistorySize, RoomShaper, Opts, QueueType])
...
 
+%% This is a top level call to normal_state to kill the muc room during any event.
+%% Another hack we should remove once we have the process that sweeps empty mucs
+normal_state(Event, StateData) ->
+	{_, _, NewStateData} = normal_state(Event, StateData, true),
+	close_room_without_occupants(NewStateData).
 normal_state({route, <<"">>,
 	      #message{from = From, type = Type, lang = Lang} = Packet},
-	     StateData) ->
+	     StateData, _) ->
```
matlegit: What's the `_` mean?

matlegit:
👍

src/mod_muc_room.erl
mpope9: It means ignore this function parameter.  Its _real_ hacky, but we'll rip it out.

src/mod_muc_room.erl
```diff
@@ -156,9 +156,14 @@ init([Host, ServerHost, Access, Room, HistorySize, RoomShaper, Opts, QueueType])
     add_to_log(room_existence, started, NewState),
     {ok, normal_state, NewState}.
 
+%% This is a top level call to normal_state to kill the muc room during any event.
+%% Another hack we should remove once we have the process that sweeps empty mucs
+normal_state(Event, StateData) ->
+	{_, _, NewStateData} = normal_state(Event, StateData, true),
+	close_room_without_occupants(NewStateData).
```
zgarbowitz: So is this essentially wrapping the call to normal_state so that we also call close_room_without_occupants after its called?

zgarbowitz: And if so, why do this as opposed to calling close_room_without_occupants at the end of the existing normal_state method?

mpope9: There are like 5 or 6 normal_state methods, and some don't save their state in a variable and simply return it.  So its easier to capture it here adjust it.

zgarbowitz:
👍

PR #81, opened 2019-03-04
[ES-1545] SQL migration file to upgrade ejabberd to 19.02

mpope9:
SQL migration for the upgraded version of ejabberd.  Ran locally without error.  The only issue is that this'll run an alter table statement on prod with 3.4m rows.  Might need to communicate with CA to have the mods announce a planned chat outage? Or maybe just pop it off....

Note, should we consider making a read replica for ejabberd?  We'd probably have to do some heavy modification to support it.

@Tdavis22 
@zgarbowitz

### DIFF
--- a/sql/2019-03-04-17-to-19-upgrade.sql
+++ b/sql/2019-03-04-17-to-19-upgrade.sql
@@ -1,0 +1,71 @@
+/**
+  * This is a SQL migration to upgrade from version 17.07 -> 19.02, sequentially from the upgrade guide
+  * found here: https://docs.ejabberd.im/admin/upgrade/#specific-version-upgrade-notes
+  *
+  **/
+
+/**
+  * 17.09 -> 17.11
+  **/
+DROP INDEX i_username ON archive;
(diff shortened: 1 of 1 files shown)


zgarbowitz:
👍

Tdavis22:
👍

nynhex:
@mpope9 

if you're worried about downtime, why not just hotswap the system?

or better yet, since it's only 3.4M rows actually you could do this a bit differently
3 different altar statements (like the altar reference?) 👿 
- one to drop secondary indices
- one to alter columnar data
- one to replace incides

this is mostly cpu bound and iops/io bound so chunking it out like this will be less obtrusive

not sure if you're on straight mysql or percona but percona would handle this faster as long as your underlying system (assuming rds or aurora) has IOPS pumps.

hell, in fact that might be a good idea to do prior to any major migration considering the weight of the table.  boost iops temporarily on rds/aurora, let the maintenance change happen, monitor closely, pull the trigger.

or...
just tell users you'll be down 🤷‍♂️

PR #82, opened 2019-03-06
[ES-1544] Adding option for send_stanza for online rooms

mpope9:
Re-adding the original logic when the room exists, because when a user is in a room and the current code is implemented, it doesn't go to the rooms history and the SDK looks weird.

@Tdavis22 
@zgarbowitz

### DIFF
--- a/src/mod_admin_extra.erl
+++ b/src/mod_admin_extra.erl
@@ -1569,18 +1569,26 @@ get_offline_user(To, From) ->
 
 %% Note this doesn't filter what we are sending to it.  Don't pass along user generated
 %% messages :) (if you need to , fun the filter_packet hook!)
+%% If the room exists, send to the room.  Otherwise, don't bring the room up.
 send_stanza(FromString, ToString, Stanza) ->
     try
 	#xmlel{} = El = fxml_stream:parse_element(Stanza),
 	To            = jid:decode(ToString),
-	From          = get_offline_user(To, jid:decode(FromString)),
-	LServer       = From#jid.lserver,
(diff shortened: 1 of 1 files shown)


src/mod_admin_extra.erl
```diff
@@ -1569,18 +1569,26 @@ get_offline_user(To, From) ->
...
-	ArchivePacket = ejabberd_hooks:run_fold(muc_filter_message, LServer, Packet, 
-	                                        [spoof_muc_state(LServer, To), From#jid.user]),
-	Wrapped       = wrap(To, From, ArchivePacket, ?NS_MUCSUB_NODES_MESSAGES),
-	PacketToSend  = xmpp:set_from_to(Wrapped, To, From),
-	ejabberd_hooks:run_fold(offline_message_hook, LServer, {bounce, PacketToSend}, [])
+	case mod_muc:find_online_room() of
+		{ok, _} ->
+			From = jid:decode(FromString),
+			Pkt = xmpp:decode(El, ?NS_CLIENT, [ignore_els]),
+			ejabberd_router:route(xmpp:set_from_to(Pkt, From, To));
```
zgarbowitz: So this is what was here originally?

mpope9: It was there because it assumed that the room was running and if it wasn't, it started it.  Then it would add the message to the MUC process's active history.  So we only want to hit this path when the room is running (to store it in the history).  Otherwise, just send to the offline/archive, which will pull it into history when the sdk wants to join the room.

zgarbowitz:
👍

Tdavis22:
I stand by my opinion. 👍

PR #86, opened 2019-03-11
[ES-1616] Fixing room creation issue and killing muc on all unavailable presences

mpope9:
### The issue
The top level function call that I added here: https://github.com/skillz/ejabberd/pull/79 was a bad idea.  We added it because I was having trouble with my local setup and couldn't actually connect.  What this did was create a 'race condition' of sorts appear _way_ more frequently.  It only happens 1% of the time on prod, going to file a different bug for that issue.

### Steps to solve
Connecting locally and setting break points, I see that the SDK is tearing down TCP connections differently(?).  This causes ejabberd to broadcast a message to all mucs that were connected to in the session once the SDK is force closed.  An unavailable presence is what is being broadcasted, but it wasn't hitting the previous logic added here: https://github.com/skillz/ejabberd/pull/78 , because the user is 'offline' at this point.

### The solution
Add a call to `close_rooms_without_occupants` at the section of code that isn't getting hit.

### Testing
Creating rooms works locally.  When leaving a room, it gets torn down.
When force closing the SDK, rooms aren't around long.
Can add friends and chat as working before.

### Problems with this fix
This fix causes all muc rooms a user has joined in a session to be temporarily brought back up.  We can prevent this by only calling that fix when the user force closes the SDK, but most users (probably?) leave the SDK open causing the rooms to still exist, instead of force closing it frequently.  SOOOO its the best of both worlds, and probably isn't really that bad anyways.

@Tdavis22 
@zgarbowitz

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -969,8 +969,8 @@ process_presence(Nick, #presence{from = From, type = Type0} = Packet0, StateData
 	     drop ->
 		 {next_state, normal_state, StateData};
 	     #presence{} = Packet ->
-		 close_room_if_temporary_and_empty(
-		   close_room_without_occupants(Nick, Packet, StateData))
+		 close_room_without_occupants(
+		   do_process_presence(Nick, Packet, StateData))
 	   end;
        true ->
 		close_room_without_occupants(StateData)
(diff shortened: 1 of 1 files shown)


mpope9:
This is a much less hacky fix, but we should still add the `mod_muc_room_sweeper` eventually, as the SDK could make another change which could cause issues.  Or not, not really sure to be honest.

Tdavis22:
👍

zgarbowitz:
👍

PR #87, opened 2019-03-12
[ES-1517] Bug online send stanza

mpope9:
### The issue
When the send_stanza was called with the code we added to fix the memory leak, it didn't actually appear in an active process's room.  It would only show up in offline messages.

### The fix
Check if the room exists.  If it does, call the old code path.
This won't result in a process leak, as the new code path doesn't bring a room up, and the old code path is only active when someone is in a room.  When that person leaves then the room will be cleaned up.

### Local testing
Two devices connected locally.  Challenging someone shows a correctly rendered message in the room.

@Tdavis22 
@zgarbowitz


zgarbowitz:
👍

Tdavis22:
👍

PR #89, opened 2019-03-15
[ES-1631] Transform subscription table

mpope9:
It looks like the work to actually use the `muc_room_subscription` table has been more or less done.  So this sql is to move all of our records from the `subscription` table to the `muc_room_subscription` table.

That string is a constant value that is provided.  Several steps of string processing is done during this query.  The table will not have data being written to it while we're draining it.

@Tdavis22 
@zgarbowitz

### DIFF
--- a/sql/2019-03-13-transform-subscription-table.sql
+++ b/sql/2019-03-13-transform-subscription-table.sql
@@ -10,9 +10,10 @@ SELECT
     SUBSTRING_INDEX(room, '@', 1),
     SUBSTRING_INDEX(room, '@', -1),
     jid,
+    SUBSTRING_INDEX(jid, '@', 1),
     '[<<"urn:xmpp:mucsub:nodes:messages">>,
-      <<"urn:xmpp:mucsub:nodes:affiliations">>,<<"urn:xmpp:mucsub:nodes:subject">>,
-      <<"urn:xmpp:mucsub:nodes:config">>]',
+  <<"urn:xmpp:mucsub:nodes:affiliations">>,<<"urn:xmpp:mucsub:nodes:subject">>,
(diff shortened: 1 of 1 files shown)


zgarbowitz:
Whats the subscriber table used for?
Do we need to modify other code in order to switch from using subscription to muc_room_subscribers?
Whats the rollout strategy here?

I'm thinking its probably safer not to truncate/drop the old subscription table right away. Can we leave it around for awhile until we are confident the new muc_room_subscribers is working?

mpope9:
I believe that processone took care of removing our code that saves subscriptions.  They substituted our constraint of populating a room with subscribers from the `moc_room_subscribers` table.  

I was going to test it on QA.  Then populate staging with like 600K rows and give it a spin there.  

I can comment out the truncate and drop.

Tdavis22:
👍

zgarbowitz:
Ok. So basically when we deploy the upgrade, at that point it will switch over to the new table? And so this SQL should be run prior to deploying? And if there's a small time gap between when we run this SQL and deploy the changes, there could be a small number of subscribers we miss?

mpope9:
No, this should be ran post-upgrade.  The new table will be read/written to, and this old table will be not read/written to.  So there will be a time where the SDK has some graphical errors due to not actually having subscribers, but no real worries of dataloss.

It should be ran as quickly as possible post deploy though.

zgarbowitz:
Why should we run it post-deploy? Isn't it better to run it before so all the data is there when the table starts getting used?

mpope9:
Yeah thats probably better.  The only reason I wanted to run this after is that we'd miss some users being written to the new table.  But this is populated whenever you access a person in your friends list, so its not the worst thing

zgarbowitz:
👍

PR #90, opened 2019-03-16
[ES-1694] Adjusting the subscription migration to deal with overlap

mpope9:
Realized the old migration didn't deal with duplicates, and there is a unique index on one of the columns.  So if there is overlap then this would error out and won't run.

EDIT: What this does is noop on duplicate keys by inserting the constant value.  Using Insert Ignore causes all errors to be treated as warnings (is what I believe I read last week), and we would still want this to error in any other case.

@Tdavis22 
@zgarbowitz

### DIFF
--- a/sql/2019-03-13-transform-subscription-table.sql
+++ b/sql/2019-03-13-transform-subscription-table.sql
@@ -4,9 +4,9 @@
   **/
 
 /* Copy data. */
-INSERT INTO 
+INSERT INTO
     muc_room_subscribers (room, host, jid, nick, nodes, created_at)
-SELECT 
+SELECT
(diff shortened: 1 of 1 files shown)


Tdavis22:
Is this because we have two users that are joining the same room? Why do we even need to update on a duplicate key in that case? Couldn't we just use insert ignore?

Tdavis22:
After reading this: https://stackoverflow.com/questions/548541/insert-ignore-vs-insert-on-duplicate-key-update it seems that on duplicate update is appropriate vs insert ignore

👍

zgarbowitz:
👍

PR #91, opened 2019-03-18
[ES-1724] Removing eimp dependency

mpope9:
In the ejabberd logs, `eimp` refuses to start.  Instead of poking at it more, I asked processone if it was essential, apparently it isn't.  I'm removing it, and `mod_avatar` and `mod_http_upload`, as we don't even use them and they rely on `eimp`.  If we want to support those in the future they're in the git history, and in processone's repo.  I'm sure this will slip through the cracks during the next upgrade, so we might run into this issue again.

`eimp` is an [image manipulation library](https://github.com/processone/eimp), fyi

@Tdavis22 
@zgarbowitz


Tdavis22:
We don't support mod_avatar? Don't our users have avatars? Is that supported in some other format?

mpope9:
We use vCards, so its cool.  https://blog.process-one.net/ejabberd-17-09/

zgarbowitz:
Is anything depending on mod_avatar or mod_http_upload? Generally all for removing unused modules 👍

Tdavis22:
Hmm, seems like we'd like mod_avatar eventually

👍

PR #99, opened 2019-04-19
[ES-2223] Fixing some ct tests

mpope9:
1)  Moving the following files to the `scripts/` directory because they're cluttering the root directory.  They were all created by us.
```
100_create_default_users.sh
build.sh
create_rooms.sh
room_removal.py
```

2) From this PR the test count changes as follows:
`416 Ok, 171 Failed, 853 Skipped of 1440` at a runtime of `1303.445s`
It used to run in `4109.727s`, with alot less of them passing.  So that's nice.

3) Ran the tests using this:
```
CT_BACKENDS=mysql,mnesia rebar ct suites=ejabberd
```
Which uses only MySQL and mensia as the backends.

4) Updated this wiki to talk about how individual tests work and how to debug them:
https://github.com/skillz/ejabberd/wiki/Tests

5) What this code change does:
I never implemented the Riak and Mnesia version of the code because we only use mysql, but it still causes test errors.  So these default values are returned when those functions are called and this code just ignores them.

@dawsonz17 
@zgarbowitz 
@hacctarr


matlegit:
👍

PR #101, opened 2019-04-25
[ES-2224] Fixing more ct tests around ownership issues

mpope9:
Adding `meck` back for future tests.

Status is now:
`532 Ok, 68 Failed, 840 Skipped of 1440` in `324.374s`.

32 more tests passing.

This change was intentionally made.  Will do some local testing to confirm if this is actually what we want.


dawsonz17:
👍

PR #102, opened 2019-04-26
[ES-2224] Fixing CT tests related to persistance and 'new_room'

mpope9:
This line:
```
true = lists:member(201, Codes),
```

Checks to see if the room is a `new_room`.  WE don't care about that because we preserve rooms and never want a room thats existed before to be new, so removing this is safe because we've changed the use case.

This changes the config to persistent.
```
    case ?config(persistent_room, Config) of
	true ->
	    [104] = set_config(Config, [{persistentroom, true}], Room),
	    ok;
	false ->
	    ok
    end.
```
and the code below changes it to not persistent.

WE don't care about this, because we want our rooms to always persist, but we actually delete them.  They're data persistent but not process persistent.

`595 Ok, 3 Failed, 842 Skipped of 1440` in `282.318s`

This fixes 65 tests and shaves off more time.

@hacctarr 
@dawsonz17 
@zgarbowitz

### DIFF
--- a/test/muc_tests.erl
+++ b/test/muc_tests.erl
@@ -1586,17 +1586,10 @@ join_new(Config, Room) ->
     ct:comment("Checking if codes '110' (self-presence) and "
 	       "'201' (new room) is set"),
     true = lists:member(110, Codes),
-    true = lists:member(201, Codes),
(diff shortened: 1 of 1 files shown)


dawsonz17:
👍

PR #105, opened 2019-05-03
[ES-2257] SQL to remove bad friends requests during previous bug

mpope9:
During this P0: https://skillzinc.atlassian.net/browse/ES-2123 friends requests couldn't go through to users.  The following SQL clears this bad data and friends requests during the time of the bug being reported and resolved will be cleared out.

Ran successfully of staging and QA and two accounts who had outstanding friends requests that couldn't be accepted were able to re-send the requests and actually complete the friend request process.

@hacctarr 
@zgarbowitz 
@dawsonz17


mpope9:
when you're friends, subscription becomes 'B' and the ask becomes 'N'.  See `/src/mod_roster.erl` for decoding/encoding rules.

dawsonz17:
👍

PR #107, opened 2019-05-07
[ES-2221] Running ct through mix

mpope9:
There are a few key things that are happening in this PR:

## 1. ct Mix task
* The test tool that ejabberd uses is `ct`.  This was previously ran using:
```CT_BACKENDS=mysql,mnesia rebar ct suites=ejabberd```
* Now, what is included is a mix task that runs ct with default options that work for the current test structure, with options to override those.
* Because some erlang options are NEEDED while running this, the task can be called through `./scripts/run_mix_ct.sh`
* Will come up with a better method (possibly) than passing the build directory and iterating over the items inside of it.

 ## 2. Changing mix's build output path
* Mix used to build to `_build`.  Now it builds to `ebin`
* Not completely necessary, but I think it will help while using the make file in the future for deployments.

## 3. Coveralls
* We use `ex_coveralls` in mod_push_skillz to generate test coverage, and now we're going to use it here.  The whole point of the mix task was to have the ability to run coveralls.
* In the future, we'll actually utilize coverage reports generated through this, but for now we're going to ignore them.

## 4. Removing Elixir Tests
* We're removing the current elixir test suite.
* This was done in mainline ejabberd
* This will be revisited and the entire test structure will be changed to accommodate these through mix task arguments.

## 5. Bumping the elixir verison
* Although this is only going to be used for the tests now, we're going to adjust the build process to only use mix. 
* This means fancy new libs can be used (like new_relic).

@hacctarr 
@dawsonz17 
@zgarbowitz


dawsonz17:
👍

PR #108, opened 2019-05-08
[ES-2221] Updating mix for umbrella child

mpope9:
We are going to be using mix under the context of a child app in an umbrella project.  This PR does that, and fixes the `ct` task to reflect its new position.  

`     child_cwd = "#{cwd}/apps/ejabberd" # TODO: kind of a hack.  this should be fixed forsure.`
This is a slight hack that I don't know how to fix.  Its not the most awful thing in the world, but if we change the directory structure then it'll be a hassle.

`File.mkdir("#{child_cwd}/ebin")` is needed because ct complains that there needs to be a local ebin directory, even though it runs with it being empty.

@hacctarr 
@dawsonz17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@
 /doc/*.toc
 /doc/contributed_modules.tex
 /doc/version.tex
-/ebin/
 /ejabberd.init
 /ejabberd.service
 /ejabberdctl.example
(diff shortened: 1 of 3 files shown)


hacctarr:
So there is no built-in reference to the app root?

mpope9:
There is a way to reference the root app, using `File.cwd!`.  There doesn't seem to be a good way of finding the child app, which is the context we're looking at.

dawsonz17:
👍

PR #117, opened 2019-05-15
[ES-2614] Reverting previous auth hack re-introduciton

mpope9:
I added some code back from Ivy's previous PR that did not work.

@hacctarr 
@dawsonz17


dawsonz17:
👍

PR #118, opened 2019-05-16
[ES-2614] Adjusting sql to delete invalid roster states

mpope9:
This sql will now delete rosters from when the initial friend request P0 came in to the present day.

@hacctarr 
@dawsonz17


dawsonz17:
👍

PR #120, opened 2019-05-17
[ES-2230] Fixing mix error with config path

mpope9:
We are in a hybrid state where ejabberd can either be in an umbrella, OR be stand alone (our current official build process).

To prevent build fails for the primary build mechanism I'm adding a check that looks for the umbrella config file.  Otherwise, use the old, ejabberd config.

@hacctarr 
@dawsonz17


dawsonz17:
👍

PR #121, opened 2019-05-17
[ES-2230] Updating test .yml to use updated certfile options

mpope9:
Changing this to use more recent certfile configs and also adjusting logger to use the Elixir Logger instead of lager, as it seems tests become unhappy with just lager.

@hacctarr 
@dawsonz17


mpope9:
Since this is only test configuration changes I'm insta merging, post mortem review would be nice.

PR #122, opened 2019-05-17
[ES-1637] Adjusting cert configs to use fully qualified paths

mpope9:
Okay, the issue before is that when running from the chat-service these config values did not have the correct path, they'd look under `chat-service/ca.pem` for example, instead of `chat-service/logs/xxx/ca.pem`.  Now we are going to use the actual ca.pem and cert.pem file path and use previously utilized substitution methods to take care of this.

@dawsonz17 
@hacctarr

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -43,7 +43,7 @@ defmodule Ejabberd.MixProject do
   def application do
     [mod: {:ejabberd_app, []},
      applications: [:kernel, :stdlib, :sasl, :ssl],
-     included_applications: [:lager, :mnesia, :inets, :p1_utils, :cache_tab,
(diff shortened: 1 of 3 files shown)


mix.exs
```diff
@@ -43,7 +43,7 @@ defmodule Ejabberd.MixProject do
   def application do
     [mod: {:ejabberd_app, []},
      applications: [:kernel, :stdlib, :sasl, :ssl],
-     included_applications: [:lager, :mnesia, :inets, :p1_utils, :cache_tab,
```
hacctarr: 👍 mmmmmm.... beer

mpope9: 🍺

dawsonz17:
👍

PR #125, opened 2019-05-23
[ES-2231] Adjusting the now janky hybrid mix file

mpope9:
Now this build works both locally and in an umbrella

@dawsonz17 
@hacctarr


dawsonz17:
👍

PR #130, opened 2019-07-18
[PS-5764] Adjusting Makefile.in and mix.exs for umbrella

mpope9:
Part of the process of moving to the mix umbrella project is using Make to install configurations, but not for normal beam files.  So this separates the actions.

Also fixing a path name, as its also changed in the chat-service.

@hacctarr 
@bhickey-skillz


bhickey-skillz:
👍

PR #131, opened 2019-07-31
[PS-6075] Add server side mute/ban and a global affiliation table

bhickey-skillz:
This PR adds support for persisting global server-side mutes and bans.

Some small background is that Ejabberd has built-in support for persisted banning via the `Outcast affiliation`, and built-in support for muting (although it is not persisted) via the `visitor role`. These are both stored in the StateData of each muc_room, which is counter-intuitive to our platform. We would like bans/mutes to be persisted and global.

This was accomplished by:
1. Adding a new sql table called user_affiliation, which maps user_id -> affiliation
2. Adding an ets_cache to quickly fetch the affiliation for any user, warmed at startup from the sql table. 
3. Adding a new affiliation called `muted` so that we can persist mutes
4. Assigning a user role to `visitor` when joining a room if their affiliation is `muted`
5. When assigning affiliations, if we receive a change to `outcast` with a reason of `muted`, we actually assign to the new `muted` affiliation. This is to allow use of the new affiliation using the existing XMPP protocol.

CAS side that makes use of these changes: https://github.com/skillz/chat-administration-service/pull/86
@aghchan @arlene-enero @mpope9 @leejona16

### DIFF
--- a/sql/2019-07-26-19-user-affiliation-table.sql
+++ b/sql/2019-07-26-19-user-affiliation-table.sql
@@ -7,12 +7,12 @@ CREATE TABLE user_affiliation (
    version       BIGINT        NOT NULL,
    enabled       BIT DEFAULT 1 NOT NULL,
    user_id       BIGINT        NOT NULL,
-   affiliation   VARCHAR(255)  NOT NULL,
+   affiliation   VARCHAR(255)  NOT NULL COLLATE utf8mb4_unicode_ci,
    date_created  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
-   last_updated  DATETIME      NOT NULL,
+   last_updated  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY(id),
-   INDEX user_affiliation_id_enabled_idx (user_id, enabled),
+   INDEX user_affiliation_enabled_id_idx (enabled, user_Id),
    INDEX user_affiliation_date_created_idx (date_created)
 )
 ENGINE = InnoDB
-DEFAULT CHARSET = utf8;
+DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
--- a/sql/2019-07-26-19-user-affiliation-table.sql
+++ b/sql/2019-07-26-19-user-affiliation-table.sql
@@ -11,7 +11,7 @@ CREATE TABLE user_affiliation (
    date_created  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
    last_updated  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY(id),
-   INDEX user_affiliation_enabled_id_idx (enabled, user_Id),
+   INDEX user_affiliation_enabled_id_idx (enabled, user_id),
    INDEX user_affiliation_date_created_idx (date_created)
 )
 ENGINE = InnoDB
--- a/sql/2019-07-26-19-user-affiliation-table.sql
+++ b/sql/2019-07-26-19-user-affiliation-table.sql
@@ -1,0 +1,18 @@
+/**
+  * This is an SQL migration to allow storage of affiliations globally rather than per-muc-room.
+  * Data will be back-filled from chat_user_quarantine after the code is changed to read/write from this table.
+  **/
+CREATE TABLE user_affiliation (
+   id            BIGINT        NOT NULL AUTO_INCREMENT,
+   version       BIGINT        NOT NULL,
+   enabled       BIT DEFAULT 1 NOT NULL,
+   user_id       BIGINT        NOT NULL,
+   affiliation   VARCHAR(255)  NOT NULL,
+   date_created  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+   last_updated  DATETIME      NOT NULL,
+   PRIMARY KEY(id),
+   INDEX user_affiliation_id_enabled_idx (user_id, enabled),
+   INDEX user_affiliation_date_created_idx (date_created)
+)
+ENGINE = InnoDB
+DEFAULT CHARSET = utf8;
--- a/src/mod_muc.erl
+++ b/src/mod_muc.erl
@@ -252,6 +252,15 @@ init([Host, Opts]) ->
 %	      load_permanent_rooms(MyHost, Host, Access, HistorySize,
 %				   RoomShaper, QueueType)
       end, MyHosts),
+    % Create a new ets cache and load it with data from the user_affiliation sql table
+    CacheOpts = [{max_size, infinity}, {cache_missed, false}, {life_time, infinity}],
+    ets_cache:new(user_affiliation_cache, CacheOpts),
+    {ok, Affiliations} = Mod:get_affiliations(Host),
+    AddToCache = fun(UserAffiliation) ->
+        {UserId, Affiliation} = UserAffiliation,
+        ets_cache:insert(user_affiliation_cache, list_to_binary(integer_to_list(UserId)), Mod:decode_affiliation(Affiliation))
+    end,
+    lists:foreach(AddToCache, Affiliations),
     {ok, State}.
 
 handle_call(stop, _From, State) ->
--- a/src/mod_muc.erl
+++ b/src/mod_muc.erl
@@ -255,12 +255,11 @@ init([Host, Opts]) ->
     % Create a new ets cache and load it with data from the user_affiliation sql table
     CacheOpts = [{max_size, infinity}, {cache_missed, false}, {life_time, infinity}],
     ets_cache:new(user_affiliation_cache, CacheOpts),
-    {ok, Affiliations} = Mod:get_affiliations(Host),
     AddToCache = fun(UserAffiliation) ->
         {UserId, Affiliation} = UserAffiliation,
-        ets_cache:insert(user_affiliation_cache, list_to_binary(integer_to_list(UserId)), Mod:decode_affiliation(Affiliation))
+        ets_cache:insert(user_affiliation_cache, integer_to_binary(UserId), Mod:decode_affiliation(Affiliation))
     end,
-    lists:foreach(AddToCache, Affiliations),
+    lists:foreach(AddToCache, Mod:get_affiliations(Host)),
     {ok, State}.
 
 handle_call(stop, _From, State) ->
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -1317,12 +1317,34 @@ set_affiliation(JID, Affiliation, StateData, Reason) ->
     Room = StateData#state.room,
     Host = StateData#state.host,
     Mod = gen_mod:db_mod(ServerHost, mod_muc),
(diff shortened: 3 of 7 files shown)


sql/2019-07-26-19-user-affiliation-table.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE user_affiliation (
+   id            BIGINT        NOT NULL AUTO_INCREMENT,
+   version       BIGINT        NOT NULL,
+   enabled       BIT DEFAULT 1 NOT NULL,
+   user_id       BIGINT        NOT NULL,
+   affiliation   VARCHAR(255)  NOT NULL,
+   date_created  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+   last_updated  DATETIME      NOT NULL,
+   PRIMARY KEY(id),
+   INDEX user_affiliation_id_enabled_idx (user_id, enabled),
```
mpope9: It feels like enabled should be first, for the initial cache population?

sql/2019-07-26-19-user-affiliation-table.sql
```diff
@@ -0,0 +1,18 @@
...
+   INDEX user_affiliation_date_created_idx (date_created)
+)
+ENGINE = InnoDB
+DEFAULT CHARSET = utf8;
```
mpope9: utf8mb4_unicode_ci I believe

sql/2019-07-26-19-user-affiliation-table.sql
```diff
@@ -0,0 +1,18 @@
...
+  * Data will be back-filled from chat_user_quarantine after the code is changed to read/write from this table.
+  **/
+CREATE TABLE user_affiliation (
+   id            BIGINT        NOT NULL AUTO_INCREMENT,
+   version       BIGINT        NOT NULL,
+   enabled       BIT DEFAULT 1 NOT NULL,
+   user_id       BIGINT        NOT NULL,
+   affiliation   VARCHAR(255)  NOT NULL,
+   date_created  DATETIME      NOT NULL DEFAULT CURRENT_TIMESTAMP,
+   last_updated  DATETIME      NOT NULL,
```
mpope9: Default here too please

src/mod_muc.erl
```diff
@@ -252,6 +252,15 @@ init([Host, Opts]) ->
 %	      load_permanent_rooms(MyHost, Host, Access, HistorySize,
 %				   RoomShaper, QueueType)
       end, MyHosts),
+    % Create a new ets cache and load it with data from the user_affiliation sql table
+    CacheOpts = [{max_size, infinity}, {cache_missed, false}, {life_time, infinity}],
+    ets_cache:new(user_affiliation_cache, CacheOpts),
+    {ok, Affiliations} = Mod:get_affiliations(Host),
+    AddToCache = fun(UserAffiliation) ->
+        {UserId, Affiliation} = UserAffiliation,
+        ets_cache:insert(user_affiliation_cache, list_to_binary(integer_to_list(UserId)), Mod:decode_affiliation(Affiliation))
```
mpope9: http://erlang.org/doc/man/erlang.html#integer_to_binary-1

src/mod_muc_room.erl
```diff
@@ -1317,12 +1317,34 @@ set_affiliation(JID, Affiliation, StateData, Reason) ->
...
+            Mod:disable_affiliation(ServerHost, LUser),
+            ets_cache:delete(user_affiliation_cache, LUser);
+        _ ->
+            ok
+        end;
+     _ ->
+        case ets_cache:lookup(user_affiliation_cache, LUser) of
+        {ok, ExistingAffiliation} ->
+            Mod:disable_affiliation(ServerHost, LUser),
+            Mod:insert_affiliation(ServerHost, LUser, list_to_binary(atom_to_list(NewAffiliation))),
```
mpope9: http://erlang.org/doc/man/erlang.html#atom_to_binary-2

src/mod_muc_room.erl
```diff
@@ -1953,6 +1976,14 @@ add_new_user(From, Nick, Packet, StateData) ->
 		  {error, Err}
 	  end;
       {_, _, _, Role} ->
+      RoomJID = StateData#state.jid,
+      To = jid:replace_resource(RoomJID, Nick),
+      case Role of
+      visitor ->
+        set_role(To, visitor, StateData);
```
mpope9: I kinda don't understand what this dose, you're not assigning the state data here

src/mod_muc_sql.erl
```diff
@@ -236,6 +238,27 @@ set_nick(LServer, Host, From, Nick) ->
 	end,
     ejabberd_sql:sql_transaction(LServer, F).
 
+-spec decode_affiliation(Arg :: binary()) -> atom().
```
mpope9: 👏

src/mod_muc_sql.erl
```diff
@@ -236,6 +238,27 @@ set_nick(LServer, Host, From, Nick) ->
...
+decode_affiliation(<<"owner">>) -> owner;
+decode_affiliation(<<"member">>) -> member;
+decode_affiliation(<<"outcast">>) -> outcast;
+decode_affiliation(<<"muted">>) -> muted;
+decode_affiliation(_) -> none.
+
+disable_affiliation(_Host, _LUser) ->
```
mpope9: Why are you ignoring the value if you're using it below?

mpope9: In several places it looks

src/mod_muc_sql.erl
```diff
@@ -245,6 +268,17 @@ set_affiliations(_ServerHost, _Room, _Host, _Affiliations) ->
 get_affiliation(_ServerHost, _Room, _Host, _LUser, _LServer) ->
     {error, not_implemented}.
 
+get_affiliations(_Host) ->
```
mpope9: Should this function hit the cache first?

src/mod_muc_room.erl
```diff
@@ -1317,12 +1317,34 @@ set_affiliation(JID, Affiliation, StateData, Reason) ->
...
+    LUser = JID#jid.luser,
+    NewAffiliation = case {Affiliation, Reason} of
+    {outcast, <<"muted">>} ->
+        muted;
+    _ ->
+        Affiliation
+    end,
+    case NewAffiliation of
+    none ->
+        case ets_cache:lookup(user_affiliation_cache, LUser) of
```
mpope9: I believe that this should be abstracted into get_affiliation and just call get_affiliation.

src/mod_muc_sql.erl
mpope9: Pattern matching ftw, you can probably use something similar instead of the atom_to_binary stuff (its safer even though we're pretty sure this is innocuous code)

mpope9: and constant time, even though I'm sure the c code underneath is probably

mpope9:
_maybe_ we should consider trying to throw up a PR into ejabberd mainline?  It'd be nice to give back.

src/mod_muc_room.erl
mpope9: I think previous caches are 'hidden in this way

bhickey-skillz: Good point

src/mod_muc_sql.erl
bhickey-skillz: I copied variable names from the unimplemented functions, which all had _ in front of it. I thought it was a convention for this module but it was just cause they weren't used :D

src/mod_muc_sql.erl
bhickey-skillz: I suppose it could and fallback when it misses, but it will miss every time in the current design. This functions only used to get all enabled affiliations, which is only used once on startup. But if it was ever used in the future it'd be nice to hit the cache so I'll add that

sql/2019-07-26-19-user-affiliation-table.sql
bhickey-skillz: Makes sense because we never lookup by user_id only atm (always user_id and enabled), but we do look up by enabled only for cache population like you said

src/mod_muc_room.erl
bhickey-skillz: This is the piece that persists visitor role status on room join. If you a join a room and your default role is visitor, assign role to visitor. (When affiliation is muted, default role = visitor)

set_role does end up assigning the state data `StateData#state{users = Users, nicks = Nicks}.`

src/mod_muc_sql.erl
bhickey-skillz: Actually I'm not sure right now how to easily do this with the ets_cache (lookup all key/value pairs) and since it would have no use case I'm going to skip this unless you feel its important

bhickey-skillz:
> _maybe_ we should consider trying to throw up a PR into ejabberd mainline? It'd be nice to give back.

I would love that - not sure who its up to if we do that. Concerns I have are that it probably doesn't fit with their design choices to support global affiliations (We could actually add a column to support a list of rooms or something like that to make it work in other people's use-cases) and the use of a non-xmpp supported affiliation might not be accepted but we could try

sql/2019-07-26-19-user-affiliation-table.sql
```diff
@@ -0,0 +1,18 @@
+/**
+  * This is an SQL migration to allow storage of affiliations globally rather than per-muc-room.
+  * Data will be back-filled from chat_user_quarantine after the code is changed to read/write from this table.
+  **/
+CREATE TABLE user_affiliation (
+   id            BIGINT        NOT NULL AUTO_INCREMENT,
+   version       BIGINT        NOT NULL,
+   enabled       BIT DEFAULT 1 NOT NULL,
+   user_id       BIGINT        NOT NULL,
+   affiliation   VARCHAR(255)  NOT NULL COLLATE utf8mb4_unicode_ci,
```
mpope9: I think `text` over `varchar` is better here, due to the fact we're storing binary data and `text` is a blob

src/mod_muc_sql.erl
```diff
@@ -236,6 +238,37 @@ set_nick(LServer, Host, From, Nick) ->
...
+    ets_cache:delete(user_affiliation_cache, LUser),
+    ok.
+
+insert_affiliation(Host, LUser, Affiliation) ->
+    AffiliationBinary = encode_affiliation(Affiliation),
+    ejabberd_sql:sql_query(Host,
+        ?SQL("insert into user_affiliation "
+        "(version, enabled, user_id, affiliation, date_created, last_updated) "
+        "values (0, 1, %(LUser)d, %(AffiliationBinary)s, now(), now())")),
+    ets_cache:update(user_affiliation_cache, LUser, {ok, Affiliation}, fun() -> ok end),
```
mpope9: Maybe we shouldn't cache the affiliation of the sql query fails.  also, why is the tuple `{ok, Affiliation}` stored instead of just `Affiliation`?

bhickey-skillz: ets_cache:update requires that the update value is in the form {ok, value} : ```The Val should be in the form of {ok, V} or error. If existing value in the cache matches Val then UpdateFun is not called```

mpope9:
Also please run the test suite and make sure no tests are broken

aghchan:
👌

PR #132, opened 2019-08-13
[PS-7793] Change user_affiliation table user_id to be a string

bhickey-skillz:
All users besides CAS have user_ids that are long. The CAS userID is 'skillz-cas' and in order to set the skillz-cas user to owner globally, this table needs to be able to accept a string. 

Additionally lowering the cap on the affiliation column (30 chars) because the longest current affiliation is 'outcast' which is only 7 chars. No need for 255 chars, but want to leave room in case we need a new affiliation in the future.

I have tested locally going through full flow of changing skillz-cas affiliation through ejabberdctl and then banning/unbanning, muting/unmuting a normal user through CAS curls


src/mod_muc.erl
```diff
@@ -257,7 +257,7 @@ init([Host, Opts]) ->
     ets_cache:new(user_affiliation_cache, CacheOpts),
     AddToCache = fun(UserAffiliation) ->
         {UserId, Affiliation} = UserAffiliation,
-        ets_cache:insert(user_affiliation_cache, integer_to_binary(UserId), Mod:decode_affiliation(Affiliation))
+        ets_cache:insert(user_affiliation_cache, UserId, Mod:decode_affiliation(Affiliation))
```
mpope9: can you change this to `Nick`, its not always a user_id

sql/2019-07-26-19-user-affiliation-table.sql
```diff
@@ -6,8 +6,8 @@ CREATE TABLE user_affiliation (
    id            BIGINT        NOT NULL AUTO_INCREMENT,
    version       BIGINT        NOT NULL,
    enabled       BIT DEFAULT 1 NOT NULL,
-   user_id       BIGINT        NOT NULL,
-   affiliation   VARCHAR(255)  NOT NULL COLLATE utf8mb4_unicode_ci,
+   user_id       VARCHAR(127)  NOT NULL,
```
mpope9: since this isn't a user_id can you rename it to nick

PR #133, opened 2019-08-13
[PS-7793] Set role to visitor on outcast affiliation change

bhickey-skillz:
Locally, my XMPP client is auto kicked when affiliation is changed to outcast.

When testing on QA, I discovered that the SDK does not auto-kick, and this bug originated by users seemingly avoiding the kick_session command. 

This ensures that if the user is banned/muted, but somehow manages to stay in the room, they at least will be set to visitor and cannot talk. 

My previous solution was to only set the role on room join, but if the user is already in a room and avoids the kick, they were able to talk still when muted/banned


mpope9:
🐘

aghchan:
🆗

PR #134, opened 2019-08-14
[PS-7793] Set role to visitor when affiliation changed to outcast

bhickey-skillz:
I did not properly return the result of set_role in my last PR, as set_role does not actually do anything besides return an updated StateData.

I have found a cleaner way, however, because by default the role gets set to none when a user is outcasted. Instead, we can set to visitor at that point. That value is passed in as StateData to set_affiliation, and subsequently returned.

I was able to test this locally by commenting out the send_kickban_presence (which is respected by my XMPP client but ignored by the SDK) so that I could stay in the room and show that I am a visitor that cannot speak


mpope9:
oh yeah, much better :electron:

aghchan:
🐺

PR #135, opened 2019-08-16
[PS-7793] Fallback to old affiliation get/set for Group-chats (DMs)

bhickey-skillz:
My recent change to use global affiliations does not work correctly with DMs, which require each participant to be owner/admin in their individual rooms. Of course we don't want to globally set users to be admins in order to allow them to join DM rooms.

The solution here is to detect if we are getting/setting an affiliation for a DM, which will contain a room of the name `nick1-nick2-nickX`, then we can fallback to the old method of getting/setting affiliations inside of the StateData record for this muc-room. 

The git Dif looks a little odd, but the logic is straight forward: (edit: actually doing hide white-space changes makes it very readable)
if room name contains a `-`, use existing fallback methods
if room does not contain a `-`, use current global affiliation cache/table logic

@aghchan @mpope9 @arlene-enero

### DIFF
--- a/src/mod_muc_room.erl
+++ b/src/mod_muc_room.erl
@@ -1315,30 +1315,35 @@ set_affiliation(JID, Affiliation,
 set_affiliation(JID, Affiliation, StateData, Reason) ->
     ServerHost = StateData#state.server_host,
     Room = StateData#state.room,
-    Host = StateData#state.host,
-    Mod = gen_mod:db_mod(ServerHost, mod_muc),
-    LUser = JID#jid.luser,
-    NewAffiliation = case {Affiliation, Reason} of
-    {outcast, <<"muted">>} ->
-        muted;
-    _ ->
-        Affiliation
-    end,
-    case NewAffiliation of
-    none ->
-        case ets_cache:lookup(user_affiliation_cache, LUser) of
-        {ok, ExistingAffiliation} ->
-            Mod:disable_affiliation(ServerHost, LUser);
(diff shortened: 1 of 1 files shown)


leejona16:
Im guessing we have no tests since ejabberd is weird about tests? 🆗

aghchan:
🍳

mpope9:
I don't think this is a smart change.  We should take a step back and re-think this approach, matching on jids doesn't make sense.

bhickey-skillz:
> I don't think this is a smart change. We should take a step back and re-think this approach, matching on jids doesn't make sense.

Please provide reasoning when adding criticism, as I have no idea why you do not think this is a smart change, despite the fact that DMs clearly have a differentiating JID than global chat rooms. And as it is a p0 I am required to fix this ASAP. My understanding is that this will fix the issue and it makes sense to us atm.

mpope9:
I think it doesn't make sense because we have a definition of a MUC room and differentiating them by JID seems like a improper approach, and if we change the definition of what a game-wide MUC JID is, this change could prove difficult to debug around.  Making a migration to move the affiliations found in the muc table's blobs into the affiliations table and do look-ups from that table would be a more consistent approach, possibly even exploring expanding the local cache (or even clearing the cache entries of affiliations when a room is destroyed to prevent unnecessary records from piling up).  The issues that are present are cropping up because they specifically don't exist in that table, if I'm understanding this correctly.

PR #138, opened 2019-08-19
[PS-8252] Enable adjusting affiliation for users that have created a DM

bhickey-skillz:
Added logs at each point along the path where can_change_ra failed, and found that when a user has created a DM and is the owner, this is the point where it fails.

I have left the logs in the other fail cases in case they still fail after setting this to true.

@mpope9 @leejona16


leejona16:
🆗 wish we had an ejabberd branch build

mpope9:
Can you remove the other error message?  I'm a bit dubious, but if we can plan for a better solution that'd be great

mpope9:
🐘

PR #141, opened 2019-09-20
Refactor dockerfile

jaredpetersen:
Refactored the existing dockerfile into something that works.

Still needs ejabberd modules to be built internally, but that's another story. For now, there's just a .gitignore for that.

@thePhilGuy 
@balusarakesh


Dockerfile
```diff
@@ -1,45 +1,126 @@
...
+        --enable-mysql \
+    && make && make install
+
+# TODO Build chat service modules
+RUN mix local.hex --force \
+    && mix local.rebar --force
+
+FROM alpine AS runtime
+
+# Install required dependencies
```
thePhilGuy: If these are the same in both the builder and runner, can they be installed in the `alpine` base where you create the ejabberd user ?

jaredpetersen: The dependencies are not the exact same -- we're not using the dev versions of the non-erlang/elixir packages in the runner stage and we're not using build-base. I'd like to pare down the dev parts for erlang/elixir too but that can't happen until we do the modules because we don't know what erlang dependencies the modules will want.

thePhilGuy: Ok that makes sense thank you

gschnaubelt:
🎉 for the multistage build!

PR #142, opened 2019-09-25
Feature add ejabberd modules

jaredpetersen:
Added private ejabberd modules to the Dockerfile. Decided to switch to a different image base since the modules required a specific elixir/erlang version that does not exist in the packages. Could have installed those manually but it doesn't make much sense when a suitable base already exists.

Also removed the linked README situation. Now it's just one `README.md`.

@balusarakesh 
@thePhilGuy


Dockerfile
```diff
@@ -17,36 +18,20 @@ RUN apk upgrade --update musl \
...
-        file \
-        curl \
-        wget \
     && rm -rf /var/cache/apk/*
 
-# Install rebar
-WORKDIR /
-RUN wget https://s3.amazonaws.com/rebar3/rebar3 -O /usr/bin/rebar3
-RUN chmod +x /usr/bin/rebar3
+# Install rebar3
```
thePhilGuy: If you're not installing rebar3 I prefer deleting lines to commenting them out

Dockerfile
```diff
@@ -55,10 +40,10 @@ RUN wget https://github.com/processone/iconv/archive/${ICONV_VERSION}.tar.gz \
 WORKDIR /tmp/iconv/iconv-${ICONV_VERSION}
 RUN rebar3 compile
```
thePhilGuy: How does this work if you've commented out the rebar3 installation above ? Is it part of `elixir:1.8-otp-22-alpine` ?

Dockerfile
jaredpetersen: Oops, sorry!

Dockerfile
jaredpetersen: Yup, that exists on the base image.

PR #143, opened 2019-09-27
Feature add kubernetes

jaredpetersen:
Added kubernetes manifests for ejabberd/chat service.

Chose a single replica deployment over a statefulset because we'll be setting up the mnesia database as an external resource in a future pull request. It's not worth our time to temporarily create a statefulset and then completely shift over to deployment later.

Added a script that automatically creates users in ejabberd if you pass the information in via environment variable. Format is `myuser@domain:mypassword anotheruser@anotherdomain:mypassword`.

Confirmed that an XMPP client (Spark) on the desktop was able to connect to the chat service running in minikube. You just have to go to advanced setting, uncheck "Automatically discover host and port", and then put in the host and port that `minikube service` gives you. That host and port needs to point to the c2s port. After that, username is `admin`, password is `password` -- we can change this later.

@thePhilGuy

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# Start chat service in the background
+exec ${EJABBERDCTL} "foreground" &
+child=$!
+${EJABBERDCTL} started
+
+# Run post startup scripts
+${EJABBERD_HOME}/scripts/create_users.sh
+
+wait $child
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 .git
 .win32
 .examples
+kubernetes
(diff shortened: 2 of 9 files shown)


scripts/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# Start chat service in the background
```
thePhilGuy: Why are you starting it in the background ?
I assumed we would want ejabberd to log to stdout so we see its logs with `kubectl logs` and also so they would automatically get shipped to logz.io

jaredpetersen: Logs still get sent to stdout.

It's weird, man.

thePhilGuy: Discussed offline, this still logs to `stdout` :+1:

PR #144, opened 2019-10-01
[ES-4115] Added configuration overlays to the kubernetes manifests

jaredpetersen:
Added configuration overlays by leveraging configMapGenerator from kustomize. This solves the problem of making sure a change to the configmap triggers a rolling update and allows us to overlay files on top of each other for different environments.

Had to modify the Dockerfile as well, as it turns out that ejabberd doesn't like where I put the ejabberd modules. It wants those in the home directory of the user that's running the node. We may be able to change that later via configuration options but I didn't see a quick and way to do so.

@thePhilGuy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,33 +52,33 @@ RUN ./autogen.sh \
     && make && make install
 
 # Build chat service modules
-RUN mkdir -p /opt/chat-service/.ejabberd-modules/sources
+WORKDIR /home/ejabberd
+RUN mkdir -p .ejabberd-modules/sources
 RUN mix local.hex --force \
     && mix local.rebar --force
 
 # Module push skillz
-WORKDIR /opt/chat-service/.ejabberd-modules/sources
+WORKDIR /home/ejabberd/.ejabberd-modules/sources
 COPY .ejabberd_modules/sources/mod_push_skillz mod_push_skillz
-WORKDIR /opt/chat-service/.ejabberd-modules/sources/mod_push_skillz
+WORKDIR /home/ejabberd/.ejabberd-modules/sources/mod_push_skillz
 RUN mix deps.get \
     && mix module_install ModPushSkillz
 
 # Module pottymouth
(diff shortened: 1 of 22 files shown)


Dockerfile
```diff
@@ -101,9 +101,11 @@ COPY --from=builder /usr/local/lib/erlang/lib/p1_utils_iconv-* /usr/local/lib/er
 WORKDIR /opt/chat-service
 COPY --from=builder /opt/chat-service .
 COPY --from=builder /tmp/chat-service/scripts scripts
+WORKDIR /home/ejabberd/.ejabberd-modules
+COPY --from=builder /home/ejabberd/.ejabberd-modules .
 
-USER ejabberd
 WORKDIR /home/ejabberd
+USER ejabberd
```
jaredpetersen: Order doesn't actually matter here. Must have changed it up when I was figuring out where to put the modules. Going to leave this as is, since it's going to take another 15 minutes to build the image and republish the image if I change it back.

kubernetes/base/config/modules.yml
```diff
@@ -0,0 +1,70 @@
...
+  mod_privacy: {}
+  mod_vcard:
+    search: true
+  ModBeamStats: {}
+  mod_pottymouth:
+    blacklists:
+      default: /home/ejabberd/.ejabberd-modules/sources/mod_pottymouth/blacklist/blacklist_en.txt
+      en: /home/ejabberd/.ejabberd-modules/sources/mod_pottymouth/blacklist/blacklist_en.txt
+  ModPushSkillz:
+    rabbit_url: "amqp://skillz-user:REDACTED@rabbit.qa.skillz.com:5672/skillz"
```
thePhilGuy: I think ejabberd fails to start if it can't reach the rabbit server for mod push skillz. Can this module be disabled just in base ?

jaredpetersen: Base is basically considered dev, so we should only be using that for local development. 

Going to take you up on this being optional, since getting this to work with our local development stack is a much larger issue.

PR #146, opened 2019-10-11
[ES-4127][Kubernetes/Vault] Pull secrets from Vault

thePhilGuy:
This configures the ejabberd pod to pull secrets from the vault running in the same kubernetes cluster.
It uses ejabberd configuration macros to fill out the parts of the config that need secrets.

There is another PR coming soon :tm:  with the external tf config map to actually connect to the right MySQL DB but I figured this could already get some review.

@balusarakesh @jaredpetersen


kubernetes/base/deployment.yaml
```diff
@@ -1,22 +1,85 @@
...
+          mountPath: /home/vault
+        env:
+        - name: VAULT_ADDR
+          value: http://vault.default:8200 # Vault is currently deployed in the default namespace
+        args: [ "agent", "-config=/etc/vault/vault-agent-config.hcl" ]
+
       containers:
-      - image: docker.dev.skillz.com:5000/ejabberd:latest
-        name: chat-service
+      # consul-template connects to vault and populates templates with the values of secrets
```
jaredpetersen: Everything that wants to use secrets needs this sidecar? That's a bummer.

kubernetes/base/kustomization.yaml
```diff
@@ -1,23 +1,28 @@
 namespace: server
 resources:
+- deployment.yaml
```
jaredpetersen: Is it okay if the deployment goes before the namespace declaration?

kubernetes/base/config/ejabberd.yaml
```diff
@@ -1,26 +1,29 @@
 language: "en"
 allow_contrib_modules: true
 
+# Secret Macros
+include_config_file: /opt/chat-service/secrets/secrets.yaml
```
jaredpetersen: Definitely confirm that renaming the files works -- last I checked, ejabberd really wanted `.yml`, at least for the base configuration file.

kubernetes/base/deployment.yaml
```diff
@@ -1,22 +1,85 @@
...
   labels:
     app: chat-service
+
 spec:
   replicas: 1
+
   selector:
     matchLabels:
       app: chat-service
+
```
jaredpetersen: Not super common to see newlines in kubernetes manifests, but I'll allow it 😉

kubernetes/base/deployment.yaml
thePhilGuy: No this is just one possible implementation that worked for me without modifying the ejabberd source.
There are plenty of other ways to get secrets from vault and I would actually encourage connecting to vault directly from the application instead of rendering templates so secrets are always kept in memory:
https://www.vaultproject.io/api/libraries.html
https://spring.io/guides/gs/vault-config/

PR #147, opened 2019-10-17
[ES-4116] Auto-clustering in kubernetes with headless service

thePhilGuy:
This allows ejabberd to come up in a statefulset with predictable DNS and automatically create a cluster.
I have tested that ejabberd continues to automatically cluster both on the initial cluster creation and on subsequent rolling updates.

This also pulls information from the external configmap created in https://github.com/skillz/infra/pull/184 and cleans up the overlay for the QA environment.

@jaredpetersen @balusarakesh please review


kubernetes/base/config/modules.yaml
```diff
@@ -10,13 +10,13 @@ modules:
...
   mod_vcard:
     search: true
-  ModBeamStats: {}
-  mod_pottymouth:
-    blacklists:
-      default: /home/ejabberd/.ejabberd-modules/sources/mod_pottymouth/blacklist/blacklist_en.txt
-      en: /home/ejabberd/.ejabberd-modules/sources/mod_pottymouth/blacklist/blacklist_en.txt
-  ModPushSkillz:
-    rabbit_url: "amqp://RABBITMQ_AUTH@rabbit.qa.skillz.com:5672/skillz" # Populated by a secret macro
+  # ModBeamStats: {}
```
jaredpetersen: Pretty sure we still want these modules

kubernetes/base/stateful-set.yaml
```diff
@@ -1,51 +1,37 @@
...
 metadata:
   name: chat-service
   labels:
     app: chat-service
-
 spec:
-  replicas: 1
-
+  replicas: 3
+  serviceName: $(SVC_NAME_INTERNAL)
```
jaredpetersen: Not sure it's really worth breaking out the variable spec of kustomize to avoid duplication in two places.

kubernetes/base/stateful-set.yaml
```diff
@@ -79,17 +64,39 @@ spec:
...
+            fieldRef:
+              fieldPath: metadata.labels['app']
+        - name: EJABBERD_INTERNAL_SVC
+          value: $(SVC_NAME_INTERNAL)
+        - name: EJABBERD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: ERLANG_COOKIE
+          value: DOYOULIKEIT
```
jaredpetersen: I like it 😄

kubernetes/base/stateful-set.yaml
```diff
@@ -79,17 +64,39 @@ spec:
...
+        - name: EJABBERD_K8S_AUTOCLUSTER
+          value: 'true'
+        - name: NUM_K8S_REPLICAS
+          value: '$(NUM_REPLICAS) '
```
jaredpetersen: These values are coming from configmaps. Why not use:
```
valueFrom:
  configMapKeyRef:
    # The ConfigMap containing the value you want to assign to SPECIAL_LEVEL_KEY
    name: chat-service-config
    # Specify the key associated with the value
    key: service_name_internal
```

thePhilGuy: I could do it that way, but since Im using the variables in more places than in `env:` definitions and container arguments, I think it's a little cleaner to use kustomize variables like this instead.

kubernetes/base/config/modules.yaml
thePhilGuy: Currently disabled in base so ejabberd can be started on its own ( without rabbitmq or mysql ). I can re-add modbeamstats I think.

kubernetes/base/stateful-set.yaml
thePhilGuy: It turns out that I was unable to use `$(NUM_REPLICAS)` the variable to set the actual number of replicas in this stateful set.
I also could not find a way to set a number of replicas in a statefulset from values in a configmap.
I've changed it back to the pattern where the `StatefulSet.chat-service.spec.replicas` is the source of truth for how many pods to run, and this `NUM_REPLICAS` variable is a reference to that value.

kubernetes/base/stateful-set.yaml
thePhilGuy: Made the name of the service a variable in response to Meidad's feedback on https://skillz.slack.com/archives/G0HECCC0P/p1571242223029400?thread_ts=1571177962.029100&cid=G0HECCC0P ( did not like the hardcoded chat-service-internal in there ).

kubernetes/base/config/ejabberd.yaml
```diff
@@ -1,8 +1,8 @@
 language: "en"
 allow_contrib_modules: true
 
-# Secret Macros
-include_config_file: /opt/chat-service/secrets/secrets.yaml
+# # Secret Macros
```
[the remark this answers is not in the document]
thePhilGuy: oh thanks I thought I had actually deleted this

PR #148, opened 2019-10-22
[PS-4816] Speed up MAM query by removing the use of count(*) on archive table

bhickey-skillz:
Hardcoding the result of the count(*) to 50. 

This breaks the ability to do ResultSet paging on MAM commands, however we do not make use of MAM for any purpose and there is no reason to support getting the middle of the archive for our use case.

count(*) on a table with 7m rows was taking over 12s on my local. In prod this table has >=16m rows and will be even worse.

By hardcoding it to 50, the query is now more like ~10ms.

We only query from the lastPage from CAS, and usually with a limit of 5. 

The value of 50 is mostly arbitrary but it allows for us to grab a larger limit later if necessary. 

@aghchan @TomPeng19971208


TomPeng19971208:
👍

aghchan:
👌

PR #150, opened 2019-11-19
[PS-12438] No longer remove offline messages on new presence

bhickey-skillz:
We have been assuming that "offline" messages function as "unread" messages. By design, when you first connect to ejabberd, all offline messages are cleared in the DB. This gives you time to display the fact that you have offline messages, and the next time you login they are gone. 

We want them to act as "unread" messages, which means that we do not clear them until explicitly told by the SDK that the messages have been viewed.

The SDK asks us to remove offline messages at the time of joining a room - DMs, group chats, etc. 

This will now actually act as "unread" messages as we expect!

@aghchan @TomPeng19971208

### DIFF
--- a/src/mod_offline_sql.erl
+++ b/src/mod_offline_sql.erl
@@ -68,7 +68,7 @@ store_message(#offline_msg{us = {LUser, LServer}} = M) ->
     end.
 
 pop_messages(LUser, LServer) ->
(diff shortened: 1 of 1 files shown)


aghchan:
👌

TomPeng19971208:
👍

PR #151, opened 2019-12-06
[ES-4122][ejabberd-metrics] fixed syntax with ejabberd metrics for prometheus

balusarakesh:
fixed syntax with ejabberd metrics for prometheus

### DIFF
--- a/kubernetes/base/kustomization.yaml
+++ b/kubernetes/base/kustomization.yaml
@@ -9,8 +9,10 @@ resources:
 - service-external.yaml
 - service-internal-discovery.yaml
 - chat-alertrule.yaml
-- configmap.yaml
 configMapGenerator:
+- name: ejabberd-statsd-mapping
+  files:
+  - config/statsd-mapping.conf
 - name: chat-service-config
   literals:
   - service_name_external=chat-service-external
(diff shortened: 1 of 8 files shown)


kubernetes/base/statefulset.yaml
```diff
@@ -17,6 +17,9 @@ spec:
     spec:
       serviceAccountName: chat-service
       volumes:
+      - name: statsd-mapping
```
thePhilGuy:
```suggestion
      # Configuration files for the statsd_exporter sidecar
      - name: statsd-mapping
```
Although, where does this configmap come from ?

balusarakesh: I forgot to push the changes, pushing them now

kubernetes/base/kustomization.yaml
```diff
@@ -8,6 +8,8 @@ resources:
 - service-account.yaml
 - service-external.yaml
 - service-internal-discovery.yaml
+- chat-alertrule.yaml
+- configmap.yaml
```
thePhilGuy: Is there any reason not to create this configmap with the format used below ?
```
- name: ejabberd-statsd-mapping
  files:
  - config/statsd-mapping.conf
```
If not, let's follow the same pattern we're using for our other configmaps. If there is let's at least rename this file to statsd-mapping-configmap because it's not the only configmap that we create for ejabberd

PR #154, opened 2020-02-11
[PS-15768] Whitelist rooms_unused_destroy endpoint

bhickey-skillz:
I had previously whitelisted using puppet-ejabberd (still necessary for production). However, we use Kubernetes now in our QA/Staging environment, which no longer uses puppet.

Instead we need to whitelist via the Kubernetes configs

@aghchan @gschnaubelt


aghchan:
💯

gschnaubelt:
🎸

PR #170, opened 2022-01-07
PS-54792: Improved get_subscribed_rooms Query

dsills-skillz:
This is an equivalent query for get_subscribed_rooms that allows the MySql query engine more freedom in generating execution plans and results in fewer reads and CPU cycles for this query. In testing, the former query read ~15k rows per query and this new one reduces that to only 2 rows, which should help the scalability of this query.


bhickey-skillz (approved):
🍌

PR #172, opened 2022-01-11
Add infinite look back to the new get_subscribed_rooms query

dsills-skillz:
Add infinite look back to the new get_subscribed_rooms query -- in testing found no performance degredation against staging DB.


bhickey-skillz (approved):
🍌

PR #174, opened 2022-01-11
Use updated git url for base64url dep

bhickey-skillz:
```
ERROR: sh(git clone -n git://github.com/dvv/base64url.git base64url)
failed with return code 128 and the following output:
Cloning into 'base64url'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

This dep is relying on the git protocol. We should instead query using https like the other dependencies by forcing it here


jgoldin-skillz (approved):
🖲️

PR #177, opened 2022-01-11
Use updated git url and tag for base64url dep in Rebar and Mix (#176)

bhickey-skillz:
* Use updated git url for base64url dep in Rebar

* Update rebar.config

* Update mix.exs


dsills-skillz (approved):
:+1:

PR #178, opened 2022-01-14
PS-55268: Add Read Replica And SQL Cache Functionality into Ejabberd

dsills-skillz:
Create new optional config, eg: 
sql_secondary_servers: ["mysql.read.replica.server.address"]

Created code in ejabberd to allow us to place the atom `secondary` in any sql_query commands we want to in ejabberd and it will use the secondary server/database. If there are multiple secondary servers, one will be picked randomly. There is also an `any` atom that can be used to randomly pick from the primary server and the list of secondary servers to spread the load between primary and secondaries evenly. If no atom is supplied, then `primary` will be used, which keeps the code backwards compatible. If `secondary` or `any` is supplied and no secondary servers are configured, then it will just use the primary server.

Testing (uses config/ejabberd.yml and you must have a local mysql instance with username, password, and database all called 'ejabberd'):
iex -S mix
:ejabberd_sql.sql_query(<<"localhost">>, "select * from muc_room_subscribers", :primary)
:ejabberd_sql.sql_query(<<"localhost">>, "select * from muc_room_subscribers", :secondary)
:ejabberd_sql.sql_query(<<"localhost">>, "select * from muc_room_subscribers", :any)
:ejabberd_sql.sql_query(<<"localhost">>, "select * from muc_room_subscribers")


src/mod_mam_sql.erl
```diff
@@ -248,6 +247,8 @@ export(_Server) ->
                     end,
                 BarePeer = jid:encode(jid:tolower(jid:remove_resource(Peer))),
                 LPeer = jid:encode(jid:tolower(Peer)),
+								ejabberd_sql_sup:invalidate_subscribed_rooms(SUser, Host),
+								ejabberd_sql_sup:invalidate_subscribed_rooms(LPeer, Host),
```
bhickey-skillz, line 251: nit this spacing seems off

bhickey-skillz (approved):
🍌

PR #203, opened 2022-02-09
PS-56912: Prevent Incoming Friend Requests to Moderators

dsills-skillz:


src/mod_block_incoming.erl
```diff
@@ -0,0 +1,298 @@
...
+
+check_subscription(From, To) ->
+  mod_roster:is_subscribed(From, To)
+.
+
+need_check(Pkt) ->
+  To = xmpp:get_to(Pkt),
+  From = xmpp:get_from(Pkt),
+
+  %% IsToAnotherUser:
```
bhickey-skillz, line 177: it might be more clear to declare these as variables and then string them together instead of commenting

src/mod_block_incoming.erl
```diff
@@ -0,0 +1,298 @@
...
+      end;
+    false -> allow
+  end
+.
+
+check_subscription(From, To) ->
+  mod_roster:is_subscribed(From, To)
+.
+
+need_check(Pkt) ->
```
bhickey-skillz, line 173: in both cases the call to need_check then calls check_subscription, we could combine this into something like... IsToBlockingUserAndNotSubscribed?

bhickey-skillz: Oh i see its following the naming in mod_block_strangers

config/dev-modules.yaml
```diff
@@ -3,6 +3,9 @@
...
...6581865@chat.skillz.com", "41241278@chat.skillz.com", "14462266@chat.skillz.com", "33411470@chat.skillz.com", "29450014@chat.skillz.com", "32704697@chat.skillz.com", "15643249@chat.skillz.com", "15550929@chat.skillz.com", "20842529@chat.skillz.com", "48178115@chat.skillz.com", "62393530@chat.skillz.com", "37172470@chat.skillz.com", "19925798@chat.skillz.com", "68727473@chat.skillz.com", "18303590@chat.skillz.com", "47350879@chat.skillz.com", "71665163@chat.skillz.com", "25325510@chat.skillz.com", "47907443@chat.skillz.com", "21710348@chat.skillz.com", "25339760@chat.skillz.com", "12959623@chat.skillz.com", "41438543@chat.skillz.com", "41481919@chat.skillz.com", "72621740@chat.skillz.com", "16242034@chat.skillz.com", "61225504@chat.skillz.com", "18456864@chat.skillz.com", "14688690@chat.skillz.com", "8817662@chat.skillz.com", "38566579@chat.skillz.com", "11338313@chat.skillz.com", "38596512@chat.skillz.com", "46910420@chat.skillz.com", "47060194@chat.skillz.com", "59207162@chat.skillz.com", "18051462@chat.skillz.com", "25000144@chat.skillz.com", "22405075@chat.skillz.com", "18099391@chat.skillz.com", "66611346@chat.skillz.com", "66361397@chat.skillz.com", "12877394@chat.skillz.com"]
```
bhickey-skillz, line 7: I'm thinking we don't need to store the jid and can just be the "nick" and append the host in code since it will always be the same env host

Would there be any limitations to storing super long lists in this format? I know we don't expect this list to grow massively and this is a cool way to easily refresh the list

src/mod_block_incoming.erl
```diff
@@ -0,0 +1,298 @@
...
+%% hook handlers
+-export([filter_packet/1, filter_offline_msg/1, filter_subscription/2]).
+
+-include("ejabberd_commands.hrl").
+-include("logger.hrl").
+-include_lib("xmpp/include/xmpp.hrl").
+
+-define(HOOK_PRIORITY, 25).
+
+-define(COMMAND_TIMEOUT, timer:seconds(30)).
```
bhickey-skillz, line 32: This seems like it might be a long timeout when its just reading the state

bhickey-skillz: and since the fallback is safe/quick (reading from the original config), it seems like we can be okay with failing quickly if it's somehow backed up

config/dev-modules.yaml
bhickey-skillz: A followup thought, do the HTTP admin commands have any method of persisting here? Or would we be doing a 2-step of persisting it here, then running the command the update it live?

src/xmpp_stream_pkix.erl
```diff
@@ -1,271 +0,0 @@
-%%%-------------------------------------------------------------------
```
bhickey-skillz, line 1: were these files deleted on purpose?

src/mod_block_incoming.erl
dsills-skillz: I agree, I switched it to 2 seconds

src/xmpp_stream_pkix.erl
dsills-skillz: Yes -- these files are already included in the dependencies for the project so I'm not quite sure why there were here, but they prevented generic xmpp clients from connecting (locally) due to some missing functions, etc, that the dependency versions do have so I removed them in order to test. Since I did this just to test, we could leave them in to be safe. Although at some point I'd like to get the project to a state where it fully works as an ejabberd server locally for easier development but that doesn't have to be this release. Let me know what you think.

src/mod_block_incoming.erl
dsills-skillz: I agree, I switched it to 2 seconds

src/xmpp_stream_pkix.erl
dsills-skillz: Yes, they are included in the dependencies so I'm not sure why they are here as well. These versions appear to be older than what the dependencies have for them and I encountered errors when trying to connect a generic xmpp client locally so I removed them to fix it so I could test locally. I'd like to get the project to a point where you can just check it out and start it and ejabberd fully works locally, but maybe that should be a separate thing let me know what you think.

config/dev-modules.yaml
dsills-skillz: To be a generic module I think we need to keep the host since ejabberd can have multiple in general, but for Skillz purposes yes that makes sense. There could be limitations if we are iterating the whole list (or half on avg) to determine membership. I will switch to a dict for better performance. This list would remain the same but the gen server would leverage a dict for its state. The http commands will update the state (list or dict) in memory but it won't persist here. I am envisioning that we would want to have a way to quickly add / remove a mod using these commands and then later at off hours we'd generate the list and set it in this config with a publish so it's persisted.

src/mod_block_incoming.erl
dsills-skillz: I switched need_check to is_blocking to be more clear since it is ultimately checking if the user is blocking incoming requests.

PR #208, opened 2022-02-28
[PS-58627] Mnesia-based Cache Module

dsills-skillz:
While currently unused, this mnesia-based cache could be leveraged in the future if we ever need to cache SQL queries or other function results in Ejabberd. Since it is currently unused, this PR changes nothing about how Ejabberd currently operates.


src/ejabberd_cache_mnesia.erl
```diff
@@ -0,0 +1,175 @@
...
+      end,
+      KeysToRemove
+    ),
+    {noreply, { CacheName, MaxCacheSize, lists:max([0, Size - length(KeysToRemove)]) }}
+  catch _:_ ->
+    {noreply, { CacheName, MaxCacheSize, Size }}
+  end
+.
+
+% got stop message
```
bhickey-skillz, line 80: nit probably don't need this comment

src/ejabberd_cache_mnesia.erl
```diff
@@ -0,0 +1,175 @@
...
+      {ok, Pid};
+    ignore -> ignore;
+    {error, Error} -> {error, Error};
+    Other -> Other
+  end
+.
+
+init([CacheName, MaxCacheSize]) ->
+  FullCacheName = get_cache_table_name(CacheName),
+  {ok, {FullCacheName, MaxCacheSize, get_size(FullCacheName, MaxCacheSize)}}.
```
bhickey-skillz, line 37: nit for consistency looks like mostly putting the '.' on the next line in the rest of this file

src/ejabberd_cache_mnesia.erl
```diff
@@ -0,0 +1,175 @@
...
+-export([start_link/2, get_proc_name/1]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, terminate/3, handle_info/3]).
+
+-include("logger.hrl").
+
+start_link(CacheName, MaxCacheSize) ->
+  %% sql cache table
+  mnesia:create_table(get_cache_table_name(CacheName), [{attributes, [ key, value ]}, {type, set}, {ram_copies, [node()]}, {local_content, false}]),
```
bhickey-skillz, line 22: i'm not sure this makes sense to run each time this process is created, is there another spot where mnesia schemas are setup?

dsills-skillz: the table name is dependent on the cache name, which is sent in as a parameter to the start_link function, which is why I placed it here. If we consolidated the cache tables into one spot outside of this module, then we'd have to pre-create all of the cache tables whereas IMO it's more flexible to just create the table with the process so that caches can be created anywhere by creating the process via start_link. If the table already exists, it does error out but just returns w/ a response that the table is already created.

bhickey-skillz (approved):
🍌

PR #211, opened 2022-03-08
[PS-59196] Remove from mod_block_incoming unnecessary message and offline message filtering

dsills-skillz:
Remove unnecessary message and offline message filtering. In my local testing for mod_block_incoming, we only need to auto-deny subscription requests. By removing the filters on messages, we will be much more efficient since subscription requests are a lot less common than messages. Also, export the is_blocking_user function for use in ModPushSkillz.


bhickey-skillz (approved):
🍌

PR #215, opened 2022-03-30
[PS-60397] Merge dev to master

dsills-skillz:
[PS-60397] 
Disallow messages to nonfriends (mod_block_nonfriends)
Fix bug where conference in server causes mod_block_incoming genserver to not be found
Sync blocking users (mods) across cluster on startup to increase probability of dynamically added mod jids from staying around.

[PS-60397]: https://skillzinc.atlassian.net/browse/PS-60397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/mod_block_nonfriends.erl
```diff
@@ -0,0 +1,110 @@
...
+-include("logger.hrl").
+-include_lib("xmpp/include/xmpp.hrl").
+
+-define(HOOK_PRIORITY, 0).
+
+%%%===================================================================
+%%% Callbacks and hooks
+%%%===================================================================
+start(Host, _Opts) ->
+  ejabberd_hooks:add(user_send_packet, Host, ?MODULE, filter_packet, ?HOOK_PRIORITY)
```
bhickey-skillz, line 28: nit I think we should change the name of the filter_packet function since that's a different hook and I was confused

dsills-skillz: good point! Renamed it

